PR #10, opened 2017-07-12
Update CONTRIBUTING.md to conform to the GCP template

crassirostris:
Copied from https://github.com/GoogleCloudPlatform/python-runtime/blob/master/CONTRIBUTING.md

/cc @piosz


piosz:
/lgtm
Could you please add the original source?

PR #11, opened 2017-07-13
Introduce omit-component-name flag.

loburm:
If flag is set to true then component name is removed from metrics that
have such prefix. For example etcd_version_info is going to be
represented as container.googleapis.com/etcd/version_info.

### DIFF
--- a/hack/verify-flags/known-flags.txt
+++ b/hack/verify-flags/known-flags.txt
@@ -97,6 +97,7 @@ nginx-configmap
 nonblocking-jenkins-jobs
 number-of-old-test-results
 ok-total-unready-count
+omit-component-name
 on-change
 one-time
 on-start
--- a/prometheus-to-sd/main.go
+++ b/prometheus-to-sd/main.go
@@ -53,6 +53,8 @@ var (
 		"Name of the pod in which monitored component is running.")
 	namespaceId = flag.String("namespace-id", "",
 		"Namespace name of the pod in which monitored component is running.")
+	omitComponentName = flag.Bool("omit-component-name", true,
+		"If metric name starts with the component name then this substring is removed to keep metric name shorter.")
 
 	customMetricsPrefix = "custom.googleapis.com"
 )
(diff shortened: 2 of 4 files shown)


prometheus-to-sd/main.go
```diff
@@ -53,6 +53,8 @@ var (
 		"Name of the pod in which monitored component is running.")
 	namespaceId = flag.String("namespace-id", "",
 		"Namespace name of the pod in which monitored component is running.")
+	squashComponentName = flag.Bool("squash-component-name", true,
```
piosz: omit?

prometheus-to-sd/main.go
```diff
@@ -136,6 +138,10 @@ func readAndPushDataToStackdriver(stackdriverService *v3.Service, gceConf *confi
 			glog.Warningf("Error while getting Prometheus metrics %v", err)
 			continue
 		}
+		if *squashComponentName {
+			metrics = translator.SquashComponentName(metrics, sourceConfig.Component)
+			sourceConfig.SquashComponentName(sourceConfig.Component)
```
piosz: I think we can assume that the metrics are already whitelisted in shortened form so there's no need to trim the prefix there too.

prometheus-to-sd/main.go
loburm: Renamed.

prometheus-to-sd/main.go
loburm: Done.

hack/verify-flags/known-flags.txt
```diff
@@ -144,6 +144,7 @@ sink-type
 skip-nodes-with-local-storage
 skip-nodes-with-system-pods
 source-file
+squash-component-name
```
piosz: fix this

loburm: Done.

loburm:
Tests have been added, flags file is fixed.

piosz:
/lgtm
Feel free to merge on green

Issue #12, opened 2017-07-13
event-exporter/sinks/stackdriver is flaky


loburm:
After closing and opening a PR issue have disappeared.

https://travis-ci.org/GoogleCloudPlatform/k8s-stackdriver/builds/253132704

crassirostris:
Whoops, forgot to replace a condition with a timeout. Thanks for noticing

PR #14, opened 2017-07-13
Fix race in event-exporter tests

crassirostris:
Since sink sends requests asynchronously, there's a delay between calling OnAdd method and calling Write method in another goroutine. There should have been a timeout until the latter method is actually called, without this timeout it was possible for the main goroutine to reach the check before second goroutine gets the chance to add enough object to the channel.

Fixes https://github.com/GoogleCloudPlatform/k8s-stackdriver/issues/12


piosz:
Could you please add short explanation what was the problem and how did you address it?

crassirostris:
@piosz PTAL

PR #18, opened 2017-08-01
Add pprofiler to the prometheus-to-sd.

loburm:
Enable on port 6060 of prometheus-to-sd pprof.

### DIFF
--- a/prometheus-to-sd/main.go
+++ b/prometheus-to-sd/main.go
@@ -18,6 +18,10 @@ package main
 
 import (
 	"flag"
+	"fmt"
+	"net/http"
+	_ "net/http/pprof"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


prometheus-to-sd/main.go
```diff
@@ -78,6 +80,10 @@ func main() {
 	}
 	glog.Infof("GCE config: %+v", gceConf)
 
+	go func() {
+		glog.Infof(http.ListenAndServe(":6060", nil).Error())
```
crassirostris: I suggest `Fatalf` instead of `Infof`, or at least `Errorf`

loburm: Done.

prometheus-to-sd/main.go
```diff
@@ -78,6 +80,10 @@ func main() {
 	}
 	glog.Infof("GCE config: %+v", gceConf)
 
+	go func() {
+		glog.Errorf(http.ListenAndServe(":6060", nil).Error())
```
crassirostris: You can probably do it in a following way:

`glog.Error(http.ListenAndServe(":6060", nil))`

crassirostris:
SGTM

prometheus-to-sd/main.go
loburm: Thanks.

crassirostris:
Just occurred to me. Have you though about two go processes exposing pprof data on this port in the same pod or on the same host with the host networking?

loburm:
Nice catch. Added flag to control exposed port.

crassirostris:
LGTM

PR #20, opened 2017-08-02
Add support of double and boolean values.

loburm:
If stored in the stackdriver MetricDescritor has type value DOUBLE or
BOOL, then convert value received from the prometheus to the correct
value before pushing it to the stackdriver.

### DIFF
--- a/prometheus-to-sd/translator/translator.go
+++ b/prometheus-to-sd/translator/translator.go
@@ -40,6 +40,8 @@ var supportedMetricTypes = map[dto.MetricType]bool{
 	dto.MetricType_HISTOGRAM: true,
 }
 
+const falseValueEpsilon = 0.001
+
 // TranslatePrometheusToStackdriver translates metrics in Prometheus format to Stackdriver format.
 func TranslatePrometheusToStackdriver(config *config.CommonConfig,
 	whitelisted []string,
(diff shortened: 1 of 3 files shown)


prometheus-to-sd/translator/translator.go
```diff
@@ -159,18 +170,31 @@ func translateOne(config *config.CommonConfig, name string, mType dto.MetricType
...
+func setValueBaseOnSimpleType(value float64, valueType string, point *v3.Point) {
+	if valueType == "INT64" {
+		val := int64(value)
 		point.Value.Int64Value = &val
 		point.ForceSendFields = append(point.ForceSendFields, "Int64Value")
+	} else if valueType == "DOUBLE" {
+		point.Value.DoubleValue = &value
+		point.ForceSendFields = append(point.ForceSendFields, "DoubleValue")
+	} else if valueType == "BOOL" {
+		var val = math.Abs(value) > falseValueEpsilon
```
piosz: @maciekpytel are you ok with such logic?

MaciekPytel: this should work fine for me (we use 0 for false and 1 for true)

piosz:
@loburm could you please bump the version to 0.2.1 and make a release?

PR #28, opened 2017-08-21
Add implementation of Custom Metrics API for multiple pod metrics.

kawych:


kawych:
@DirectXMan12 
Please take a look.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -115,9 +116,26 @@ func (p *StackdriverProvider) GetNamespacedMetricByName(groupResource schema.Gro
 
 // GetNamespacedMetricBySelector queries Stackdriver for metrics identified by selector and associated
 // with a namespace.
-// TODO(kawych): implement this
 func (p *StackdriverProvider) GetNamespacedMetricBySelector(groupResource schema.GroupResource, namespace string, selector labels.Selector, metricName string) (*custom_metrics.MetricValueList, error) {
-	return nil, fmt.Errorf("Selectors not supported")
+	matchingObjectsRaw, err := p.apiserverClient.Get().
```
piosz: /s/apiserverClient/kubeClient/

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
```diff
@@ -84,6 +85,49 @@ func (t *Translator) GetRespForPod(response *stackdriver.ListTimeSeriesResponse,
 	return nil, fmt.Errorf("Illegal state")
 }
 
+// GetSDReqForPods returns Stackdriver request for query for multiple pods.
+func (t *Translator) GetSDReqForPods(podList runtime.Object, metricName string) (*stackdriver.ProjectsTimeSeriesListCall, error) {
```
piosz: can we pass podList here instead of runtime.Object?

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
```diff
@@ -84,6 +85,49 @@ func (t *Translator) GetRespForPod(response *stackdriver.ListTimeSeriesResponse,
...
+	return t.service.Projects.TimeSeries.List(project).Filter(filter).
+		IntervalStartTime(startTime.Format(time.RFC3339)).
+		IntervalEndTime(endTime.Format(time.RFC3339)).
+		AggregationPerSeriesAligner("ALIGN_NEXT_OLDER").
+		AggregationAlignmentPeriod(fmt.Sprintf("%vs", int64(t.reqWindow.Seconds()))), nil
+}
+
+// GetRespForPods returns translates Stackdriver response to a Custom Metric associated
+// with multiple pods.
+func (t *Translator) GetRespForPods(response *stackdriver.ListTimeSeriesResponse, podList runtime.Object, groupResource schema.GroupResource, metricName string, namespace string) (*custom_metrics.MetricValueList, error) {
```
piosz: same here

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
piosz: And in the other places which are already implemented like `GetSDReqForPod` or `GetRespForPod`.

piosz: There is a lot of common logic here and in `GetSDReqForPod`. Why not reuse it somehow? Especially one pod can be treated as a list of pods.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
kawych: Done.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
kawych: What do you mean? Do you want me to pass here the list of pod IDs?

The reasoning I had in mind was to put as much logic here as opposed to provider.go, because only the translator is unit tested. When you look at function: GetRespForPods, there is another reason for passing runtime.Object - I use there both pod names and IDs, so I would have to create some other structure to pass this as an argument.

What I don't like here is that pods/podLists are not typed. I agree it makes more sense to convert podList and pod to typed objects as soon as possible, i.e. in provider.go. I decided that here it's not worth it. WDYT?

kawych: Extracted some of it, the rest depends on the comment above.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
kawych: Which of your previous comment do you refer? I don't find it straightforward to apply any of them here.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
piosz: I want to take advantage of golang which is typed. When I see podList I expect it's indeed Pod list. Without this you can pass namespace list by mistake and you won't catch the problem.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
piosz: both;)
1. pass `podList` as type `v1.PodList`
2. share logic `GetRespForPods`

piosz:
High level comment:
You have methods:GetSDReqForPods and `GetSDReqForPods`.
The second method has to handle the case when there is only one pod on the list.
Why not just have then method `GetSDReqForPods` and call it as `GetSDReqForPods(PodList{Items: Pod[]{pod}}` when you need to call  `GetSDReqForPod`?

The same apply to `GetRespForPod` and `GetRespForPods`

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -55,7 +56,7 @@ func (c realClock) Now() time.Time {
 
 // StackdriverProvider is a provider of custom metrics from Stackdriver.
 type StackdriverProvider struct {
-	apiserverClient    rest.Interface
+	kubeClient         rest.Interface
```
piosz: BTW is there any reason why this is rest.Interface rathern than [CoreV1Client](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/core_client.go#L47)? The latter offers you native pods methods (see [here](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/core/v1/pod.go#L35) and you can avoid passing `runtime.Object` and conversion problems below.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
kawych: Done.

kawych: Done (in a way you requested).

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
kawych: 1. Done
2. This is a little different than GetReq*, because the results are different, i.e. have different type. I think that all common logic is shared via functions getMetricValuesFromResponse() and metricFor().

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
kawych: As discussed offline, I agree it's better to switch to CoreV1Client.

piosz:
Please rebase

PR #33, opened 2017-08-30
Add an empty implementation to list all events and to create a new event

erocchi:


event-adapter/cmd/provider/provider.go
```diff
@@ -19,30 +19,40 @@ package provider
...
-// ListAllEventsByNamespace gets all the events
-func (p *StackdriverProvider) ListAllEventsByNamespace(namespace string) (*types.EventValueList, error) {
+// ListAllEventsByNamespace gets all the events with the given namespace
+func (p *StackdriverProvider) ListAllEventsByNamespace(namespace string) (*api.EventList, error) {
 	return nil, fmt.Errorf("ListAllEventsByNamespace is not implemented yet.")
 }
+
+// ListAllEvents gets all the events
+func (p *StackdriverProvider) ListAllEvents() (*api.EventList, error) {
+	return nil, fmt.Errorf("ListAllEvents is not implemented yet.")
```
kawych: Note here: the returned errors should be Kubernetes AIP - compatible errors, see example definitions here: https://github.com/kubernetes-incubator/custom-metrics-apiserver/blob/master/pkg/provider/errors.go
This changed in the framework after you forked from it. Maybe a TODO?

event-adapter/pkg/apiserver/installer/context/context.go
```diff
@@ -20,9 +20,14 @@ import (
 	"k8s.io/apiserver/pkg/endpoints/request"
 )
 
-// resourceInformation holds the resource and subresource for a request in the context.
+// resourceInformation holds the resource and method for a request in the context.
```
kawych: What is method?

I think the name 'subresource' is a more abstract concept, and api machinery operates on this level of abstraction, so it would make more sense to use this word. I'm not sure if I understand all of this correctly, so please explain what is the reasoning behind calling this field 'method' and whether or not it is consistent with another field named 'resource'.

event-adapter/pkg/apiserver/installer/installer.go
```diff
@@ -250,13 +239,60 @@ func (a *EventsAPIInstaller) registerResourceHandlers(storage rest.Storage, ws *
...
-	//REGISTER LIST ALL EVENT
+	//PATH namespaces/{namespace}/events/{eventName}
+
 	namespacedListPath := scope.ParamName() + "/{" + scope.ArgumentName() + "}/events"
 	namespacedListParams := []*restful.Parameter{
 		namespaceParam,
-		resourceParam,
 	}
 
+	doc = "event with the given namespace"
```
kawych: Please be consistent with other docs. It should be in imperative form, like "list all events".

kawych:
Great, small PR. Looks good overall, just a couple of comments.

event-adapter/cmd/provider/provider.go
erocchi: Done

event-adapter/pkg/apiserver/installer/installer.go
erocchi: Done

event-adapter/cmd/provider/provider.go
kawych: Thanks. Please move the TODO to some more visible place, e.g. top of this file.

event-adapter/pkg/apiserver/installer/context/context.go
kawych: From offline discussion - `method` is a method of http request, e.g. GET or POST. It is however a little bit confusing, since it shouldn't be a part of resource specification. Can you find some more appropriate naming? For example, change `resourceInformation` struct to `requestInformation`

event-adapter/cmd/provider/provider.go
erocchi: Done

event-adapter/pkg/apiserver/installer/context/context.go
erocchi: Done

PR #39, opened 2017-09-11
Add an implementation to read from Stackdriver

erocchi:


event-adapter/adapter.go
```diff
@@ -23,15 +23,21 @@ import (
 	"k8s.io/apiserver/pkg/util/logs"
 )
 
+var (
+	maxRetrievedEvents = flag.Int("max-retrieved-events", 100, "Maximum number of events can be retrieved")
+	sinceMillis        = flag.Int64("retrieve-events-since-millis", -1, "Retrieve only events that where logged at most the given amount of seconds ago")
```
kawych: Please update a message - meaning of this flag is a little different as we discussed.

erocchi: Updated

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
 	"github.com/GoogleCloudPlatform/k8s-stackdriver/event-adapter/pkg/provider"
-	stackdriver "google.golang.org/api/monitoring/v3"
-	"k8s.io/client-go/pkg/api"
+	sd "google.golang.org/api/logging/v2"
+	api "k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/rest"
+	"time"
 )
 
+//TODO(erocchi) add implementation to write to Stackdriver
```
kawych: Maybe "Add implementation to write events to Stackdriver" would be just a little bit clearer?

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
 //TODO(erocchi) Use the Kubernetes AIP - compatible errors, see example definitions here: https://github.com/kubernetes-incubator/custom-metrics-apiserver/blob/master/pkg/provider/errors.go
 
-type sdService stackdriver.Service
-
 // StackdriverProvider implements EventsProvider
 type StackdriverProvider struct {
-	//TODO(erocchi) Define fields for this struct. It should contain what will be needed to communicate with Stackdriver.
+	restClient         rest.Interface
```
kawych: Where do you use restClient?

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
+	service            *sd.Service
```
kawych: Maybe "stackdriverService" or "sdService"? Just "service" is too ambiguous.

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
 }
 
 // CreateNewEvent creates a new event in the given namespace
 func (p *StackdriverProvider) CreateNewEvent(namespace string) (*api.Event, error) {
 	return nil, fmt.Errorf("CreateNewEvent is not implemented yet.")
 }
+
+func (p *StackdriverProvider) doRequest(filter string) (*api.EventList, error) {
+	projectID, err := gce.ProjectID()
+
```
kawych: Please remove the empty lines unless they really help readability. For example there is no point in adding empty line between sth. like "foo, err := bar()" and "if err != nil { ... }".

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
+	if err != nil {
+		return nil, fmt.Errorf("Cannot retrieve projectID. %v", err)
+	}
+
+	resource := []string{
+		"projects/" + projectID,
+	}
+
+	entriesService := sd.NewEntriesService(p.service)
```
kawych: Is there a reason why you create entriesService and newEventsTranslator each time you send a request instead of doing it a single time, for example in NewStackdriverProvider function?

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
+	translator := newEventsTranslator(p.maxRetrievedEvents)
+
+	var timeLimit time.Time
+	if p.sinceMillis < 0 {
```
kawych: Please replace this with the logic we spoke about. Feel free to reach to me if you're unsure about anything.

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
+		timeLimit = time.Now().Add((-3600) * time.Second)
+	} else {
+		timeLimit = time.Unix(0, p.sinceMillis*1e6)
+	}
+	timeFilter := "(" + filter + " AND " + "jsonPayload.metadata.creationTimestamp >= \"" + timeLimit.Format("2006-01-02T15:04:02.002Z") + "\")"
```
kawych: Please use time.RFC3339 constant here.

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
+
+	eventsList := &api.EventList{}
+	var response *sd.ListLogEntriesResponse
+
+	for redirection := 0; redirection < 1000; redirection++ {
```
kawych: Please don't use such literals (1000) in the code unless it's very clear what do they mean and where did the value come from. If it's not obvious why this value is chosen, it's better to comment the reasoning or create a constant or flag. It's easy to end up with such literals scattered all over the code, which may sometimes cause unexpected and hard to explain behavior (as this value isn't documented anywhere).

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
+		request := &sd.ListLogEntriesRequest{
```
kawych: Does it make sense to set PageSize in the request, so that Stackdriver doesn't return 1000 of entries when only 1 is requested?

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
+	return "(logName = \"projects/" + projectID + "/logs/events\" AND jsonPayload.kind = \"Event\")", nil
+
 }
 
 // CreateNewEvent creates a new event in the given namespace
 func (p *StackdriverProvider) CreateNewEvent(namespace string) (*api.Event, error) {
 	return nil, fmt.Errorf("CreateNewEvent is not implemented yet.")
 }
+
+func (p *StackdriverProvider) doRequest(filter string) (*api.EventList, error) {
```
kawych: Please find a more descriptive name. "doRequest" would be perfect if the argument was some structure representing a request. This function does more than simply executing a single request and the name doesn't say what kind of request will be sent.

event-adapter/cmd/provider/provider.go
```diff
@@ -17,44 +17,146 @@ limitations under the License.
...
+			Filter:        timeFilter,
+			OrderBy:       "timestamp desc",
+		}
+		if redirection != 0 {
+			request.PageToken = response.NextPageToken
+		}
+		response, err = entriesService.List(request).Do()
+
+		if err != nil {
+			return nil, fmt.Errorf("Impossible to get the answer:  %v", err)
```
kawych: I find the wording here a little bit unprecise (it's not "impossible", we only know that our attempt failed). Maybe "Failed request to Stackdriver: ..."?
A little bit different case below, but I still don't like the word. I'd prefer "Failed to translate the response from Stackdriver: ...".

event-adapter/cmd/provider/translator.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	sd "google.golang.org/api/logging/v2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	api "k8s.io/client-go/pkg/api/v1"
+)
+
+type eventTime struct {
```
kawych: What is the benefit of defining new struct? Can you use metav1.Time instead?

event-adapter/cmd/provider/translator.go
```diff
@@ -0,0 +1,114 @@
...
+	Timestamp metav1.Time `json:"timestamp,omitempty"`
+}
+
+type eventsTranslator struct {
+	eventsList    *api.EventList
```
kawych: What's the benefit of having eventsList here in comparison to having only a local variable in function "translateToEventList"? The same applies to the other two fields. Going further into that, do you even need eventsTranslator structure? Can you remove it?

event-adapter/cmd/provider/translator.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	return &time, nil
+
+}
+
+func (t *eventsTranslator) translateToEventList(response *sd.ListLogEntriesResponse) (*api.EventList, error) {
+
+	for _, value := range response.Entries {
+		event := value
+		newEvent, err := translateToEvent(event)
```
kawych: IUUC, the event (which is SD log entry) contains information both about event and timestamp. "translateToEvent" extracts from this entry information about event and "translateToTime" extracts information about timestamp. Is this correct? If so, the name of these functions is confusing. "translate" suggests that you convert whole entry to other structure, but instead you are just extracting a piece of information. Please fix the naming.

event-adapter/cmd/provider/provider.go
erocchi: Done

event-adapter/cmd/provider/provider.go
erocchi: Removed

event-adapter/cmd/provider/provider.go
erocchi: Done

event-adapter/cmd/provider/provider.go
erocchi: Done

event-adapter/cmd/provider/provider.go
erocchi: Moved

event-adapter/cmd/provider/provider.go
erocchi: Done

event-adapter/cmd/provider/provider.go
erocchi: Done

event-adapter/cmd/provider/provider.go
erocchi: Removed

event-adapter/cmd/provider/provider.go
erocchi: Done

event-adapter/cmd/provider/provider.go
erocchi: Done

event-adapter/cmd/provider/translator.go
erocchi: Removed

event-adapter/cmd/provider/provider.go
kawych: There is still a lot of empty lines all over this file (and in other files). Please apply this comment everywhere.

event-adapter/cmd/provider/provider.go
kawych: You can use here another constant "time.Hour"

erocchi: Done

event-adapter/cmd/provider/provider.go
kawych: Resolved

event-adapter/cmd/provider/translator.go
erocchi: I use this struct in order to decode the response. I need to access the field timestamp in the response, in this way it decodes the response in an eventTime struct and put the timestamp value in the field Timestamp. I do not know if there is a better way that allow me to access that value.

erocchi: Removed the struct, used a map instead.

PR #42, opened 2017-09-12
Update config for example

tonglil:
Use beta apis, update the image, and create a new CRB.

### DIFF
--- a/event-exporter/example/event-exporter.yaml
+++ b/event-exporter/example/event-exporter.yaml
@@ -20,7 +20,7 @@ metadata:
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
-  name: event-exporter-rb-default
+  name: event-exporter-rb-example
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
@@ -43,7 +43,7 @@ spec:
     spec:
       serviceAccountName: event-exporter-sa
       containers:
-        - name: event-exporter
-          image: gcr.io/google-containers/event-exporter:v0.1.7
-          command:
-            - '/event-exporter'
+      - name: event-exporter
+        image: gcr.io/google-containers/event-exporter:v0.1.7
+        command:
+          - '/event-exporter'
--- a/event-exporter/example/event-exporter.yaml
+++ b/event-exporter/example/event-exporter.yaml
@@ -17,19 +17,18 @@ kind: ServiceAccount
 metadata:
   name: event-exporter-sa
 ---
-apiVersion: rbac.authorization.k8s.io/v1alpha1
+apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
-  name: event-exporter-rb
+  name: event-exporter-rb-default
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: view
 subjects:
-- apiVersion: v1
-  kind: ServiceAccount
-  name: event-exporter-sa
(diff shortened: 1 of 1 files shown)


event-exporter/example/event-exporter.yaml
```diff
@@ -17,19 +17,18 @@ kind: ServiceAccount
 metadata:
   name: event-exporter-sa
 ---
-apiVersion: rbac.authorization.k8s.io/v1alpha1
+apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
 metadata:
-  name: event-exporter-rb
+  name: event-exporter-rb-default
```
tonglil: Had to create a new one, kube would not let me update the existing one.

event-exporter/example/event-exporter.yaml
```diff
@@ -44,7 +43,7 @@ spec:
     spec:
       serviceAccountName: event-exporter-sa
       containers:
-      - name: event-exporter
-        image: gcr.io/google-containers/event-exporter:v0.1.4
-        command:
-        - '/event-exporter'
+        - name: event-exporter
```
piosz: one tab less?

tonglil: Ok let me do that. I think my yaml linter automatically indented this.

tonglil: Yeah the linter is saying that it should be indented (and judging by eye it looks like it should be too).
Still want me to de-indent?

crassirostris: It doesn't matter, either way is correct in case of lists

In this particular case it's just how the whole file was formatted, why change?

event-exporter/example/event-exporter.yaml
crassirostris: Maybe append a version instead? v1beta will change to v1 at some point anyway

tonglil: Or I can append `-example`?

event-exporter/example/event-exporter.yaml
tonglil: I'll de-indent (standard yml editor complained and did it was the original reason).

event-exporter/example/event-exporter.yaml
crassirostris: Sorry, I realized I don't understand the premises

> because that was being used by kube-system.

What do you mean?

event-exporter/example/event-exporter.yaml
crassirostris: You can just change the number manually, without editing the rest of the file

event-exporter/example/event-exporter.yaml
crassirostris: Oh, do you mean it doesn't work in a cluster with event-exporter already running as part of the deployment?

tonglil: Yes, that is what I mean, because they are cluster wide (not namespaced).

event-exporter/example/event-exporter.yaml
tonglil: Sorry I was editing this in vim (which indented automatically on save).
Is it ok now, or something I should change?

crassirostris:
@tonglil Sorry about this mess with indentation :) Let's just format the file how your tool suggests, so we don't even have this problem in the future, and I will lgtm

event-exporter/example/event-exporter.yaml
crassirostris: OK, that's fine with me

crassirostris:
LGTM

crassirostris:
Thanks!

PR #51, opened 2017-09-20
Rebase Custom Metrics - Stackdriver Adapter to new versions of Kubernetes dependencies

kawych:


piosz:
@sttts would you mind taking a look?

custom-metrics-stackdriver-adapter/pkg/apiserver/installer/installer_test.go
```diff
@@ -23,7 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apiserver/pkg/endpoints/handlers"
 	"k8s.io/apiserver/pkg/endpoints/request"
-	"k8s.io/client-go/pkg/api"
+	"k8s.io/kubernetes/pkg/api"
```
sttts: this should be `k8s.io/api/core/v1`

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -28,11 +28,11 @@ import (
 	"k8s.io/metrics/pkg/apis/custom_metrics"
 
 	// Install registers the API group and adds types to a scheme.
-	_ "k8s.io/client-go/pkg/api/install"
+	_ "k8s.io/kubernetes/pkg/api/install"
```
sttts: Which scheme are you talking about here? The client-go clientset has its own scheme. Where do you need another scheme?

In general, you shouldn't depend on `k8s.io/kubernetes`.

custom-metrics-stackdriver-adapter/pkg/cmd/server/start.go
```diff
@@ -73,15 +73,15 @@ func (o CustomMetricsAdapterServerOptions) Config() (*apiserver.Config, error) {
 		return nil, fmt.Errorf("error creating self-signed certificates: %v", err)
 	}
 
-	serverConfig := genericapiserver.NewConfig(apiserver.Codecs)
-	if err := o.SecureServing.ApplyTo(serverConfig); err != nil {
+	serverConfig := genericapiserver.NewRecommendedConfig(apiserver.Codecs)
+	if err := o.SecureServing.ApplyTo(&serverConfig.Config); err != nil {
```
sttts: There is also `NewRecommendOptions` with one "big" `ApplyTo`.  That way you won't forget to initialize important fields.

sttts: compare https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/sample-apiserver/pkg/cmd/server/start.go#L40.

custom-metrics-stackdriver-adapter/pkg/apiserver/installer/apiserver_test.go
```diff
@@ -127,12 +127,10 @@ func handle(prov provider.CustomMetricsProvider) http.Handler {
 
 			Context:                reqContextMapper,
 			OptionsExternalVersion: &schema.GroupVersion{Version: "v1"},
-
-			ResourceLister: provider.NewResourceLister(prov),
 		},
 	}
 
-	if err := group.InstallREST(container); err != nil {
+	if err := group.InstallREST(container, prov, reqContextMapper); err != nil {
```
sttts: without having an overview about your code structure, the creation of another request context mapper looks suspicious. It's a global state and having a second one is usually wrong.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
```diff
@@ -21,22 +21,21 @@ import (
...
+	"github.com/golang/glog"
 	stackdriver "google.golang.org/api/monitoring/v3"
+	"k8s.io/api/core/v1"
 	apierr "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/pkg/api"
+	"k8s.io/kubernetes/pkg/api"
```
sttts: all these `k8s.io/kubernetes` imports should be replaced with corev1 from k8s.io/api or with the scheme from the client.

kawych: Done.

custom-metrics-stackdriver-adapter/pkg/cmd/server/start.go
kawych: Done

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
kawych: Removed kubernetes dependency

custom-metrics-stackdriver-adapter/pkg/apiserver/installer/installer_test.go
kawych: Done.

custom-metrics-stackdriver-adapter/pkg/apiserver/installer/apiserver_test.go
kawych: I'm not familiar with the logic that creates reqContextMapper, can we leave it as it is? If this is a bug, I think it should be fixed here (where this code is forked from): https://github.com/kubernetes-incubator/custom-metrics-apiserver/blob/master/pkg/apiserver/installer/apiserver_test.go

piosz: It think is should be fixed in both places.
cc @DirectXMan12

kawych: I agree, IF it is a bug. I think it's not, it looks to me that the request context mapper is created only here for testing purposes.

What I mean is, we don't need to block this PR on this, this is unrelated to the changes I'm making.

sttts: right, in tests this can be ok.

kawych:
@sttts 
Thank you for the review. I have addressed all comments, PTAL.

piosz:
/lgtm
Feel free to merge on green.

kawych:
Merging. @sttts If you have further comments, please add them, I'll address in next PR.

PR #59, opened 2017-10-18
Fix setup for Custom Metrics - Stackdriver Adapter in GKE

kawych:


kawych:
@fgrzadkowski 
Context: https://github.com/kubernetes/kubernetes/issues/43409

So this can be fixed by creating cluster role binding somewhere in e2e tests to make it work, it would also require another step for GKE clusters in the user guide (binding user to RBAC role cluster-admin). Alternatively we can use existing cluster role 'view', which is my preference.

fgrzadkowski:
Ok, I don't like the workaround, but I understand motivation...

PR #61, opened 2017-10-27
prometheus-to-sd: support additional labels for source

atetubou:
This CL is used to add custom labels to time series metrics scraped from pods.

Allowing custom labels makes aggregation more easy and useful on stackdriver.


atetubou:
Discussed with authors, I found that this PR is not right way for our usage. Let me close this.

Issue #52, opened 2017-09-20
Default Value Type for prometheus-to-sd


filipe-cantarelli:
The current default type is ["INT64"](https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/prometheus-to-sd/translator/translator.go#L282)

This is breaking floating point metrics such as: "custom.googleapis.com/kube-state-metrics/kube_pod_container_resource_requests_cpu_cores"

Since the variable that holds the Value is of type float64 already, I think the best candidate for default type should be "DOUBLE" instead of "INT64".

loburm:
It's a duplicate of: https://github.com/kubernetes/contrib/issues/2655

loburm:
Oh, It's in wrong repository.

It actually have a sense to leave this one.

acondrat:
I was having the same issue and the default type change seems to be a reasonable solution. 
@filipe-cantarelli 👍 

 submitted #67

ViliusS:
@loburm will this ever be fixed? This constantly bites us in the $%&.

For example, we've been running [RabbitMQ Cluster from Google Marketplace](https://github.com/GoogleCloudPlatform/click-to-deploy/blob/master/k8s/rabbitmq/README.md), which installs with prometheus-to-sd as sidecar container and configures metric descriptors as int64 values.
Now we want to monitor our RabbitMQ clients (Java applications), which uses Spring Boot GCP libraries and configures metric descriptors as double values. The application runs in the same GCP project on GKE.
Coincidently both, RabbitMQ cluster and RabbitMQ client applications, have two metrics under the same name: `custom.googleapi.com/rabbitmq/connections` and `custom.googleapi.com/rabbitmq/channels`. Since prometheus-to-sd created metric descriptors with INT64 we cannot collect these metrics from Java application. We also cannot change metric descriptor because it comes from Spring Boot and/or GCP libraries. The following error is produced:
```
failed to send metrics to Stackdriver
[...]

PR #67, opened 2017-11-14
#52 setting 'DOUBLE' as the default type

acondrat:


acondrat:
I signed it!

loburm:
@acondrat have you checked what will happen if we try to rewrite value type for already existing metric in stackdriver? Currently we already have prometheus-to-sd usages that we don't want to break.

acondrat:
@loburm looks like the valueType is not updated. I had to manually recreate the metric descriptor with  [projects.metricDescriptors/delete](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors/delete) and [projects.metricDescriptors/create](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors/create)

Simply deleting the metricDescriptor was not enough. For some reason it keeps using the old (INT64) valueType. I suspect [projects.metricDescriptors/list](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors/list) is cached.

loburm:
In such case we can't merge this PR yet, because some of the existing metrics are going to be broken. I'll try to come up with better solution.

filipe-cantarelli:
As far as I remember when I was looking at the issue, this change will only affect new metrics(metrics that don't have a metricDescriptor yet). For any old metric, the type specified in the metricDescriptor will be used instead.

loburm:
That's not true. If original metric descriptor is missing from the cache, then we are going to use the default one. That's a case for GKE metrics that are writable but not readable.

PR #68, opened 2017-11-16
Revert "Don't set custom metrics reader permission in adapter deployment"

kawych:
Reverts GoogleCloudPlatform/k8s-stackdriver#65

It triggered this issue: https://github.com/kubernetes/kubernetes/issues/55805


fgrzadkowski:
\lgtm

PR #71, opened 2017-11-27
End-to-end test for prometheus-to-sd

jkohen:
Verifies that prometheus-to-sd exports the right monitored resources to Stackdriver, and ensures the value of the time series is correct (for int64 metrics currently). Right now the focus is on monitored resources, in preparation for the integration with metadata agent, and basic communication with the Stackdriver Monitoring service. The secondary focus is to test that the included deployment file (e2e.yaml) works correctly, so that it can be used as sample configuration in the documentation.

Each execution runs in its own namespace.

Run with go test ./integration -integration -cluster-name=my-cluster


piosz:
It's huge. We will do our best to take a look, though I'm not sure how long would it take.

jkohen:
Apologies for the length. I added <400 lines of code, everything else is
under /vendor and I don't know what I could do to make it smaller. I also
don't think you need to review that.

prometheus-to-sd/integration/stackdriver.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/cenkalti/backoff"
+	monitoring "google.golang.org/api/monitoring/v3"
+)
+
+// Returns the value from a TypedValue as an int64. Floats are returned after
```
loburm: Please start the comment with function name.

prometheus-to-sd/integration/e2e_test.go
```diff
@@ -0,0 +1,121 @@
...
+	if !*integration {
+		t.Skip("skipping integration test: disabled")
+	}
+	if *clusterName == "" {
+		t.Fatalf("the cluster name must not be empty")
+	}
+	log.Printf("Cluster name: %s", *clusterName)
+	namespaceName := fmt.Sprintf("e2e-%x", rand.Uint64())
+	log.Printf("Namespace name: %s", namespaceName)
+	if err := execKubectl("create", "namespace", namespaceName); err != nil {
```
loburm: Executing kubectl command directly doesn't sound as a good idea. Why not to use methods that are provided by k8s golang lib? For example you can use:
GetClientset().CoreV1().Namespaces().Create()?

I believe that for other calls it's also possible to find methods.

jkohen:
Thank you for your review. Because we ended up going in a different direction and not going to use prometheus-to-sd, I'm not going to submit this PR. However, I am reusing the end-to-end test almost directly from what you reviewed, and I've incorporated your suggestions and feedback into my version. I really appreciate it.

PR #72, opened 2017-11-27
Code for Prometheus Gauge to Stackdriver Cumulative translation logic

rramkumar1:
Adds a new flag which takes in a comma-separated list of metric names. These are the metrics which will undergo Prometheus Gauge -> Stackdriver Cumulative translation. This entails the following:

1. The value will get a process_start_time_seconds timestamp (just like a Cumulative). 
2. The value of the Gauge will directly translate over to the Cumulative value. No extra processing is necessary.
3. This will not work for metrics being pushed to custom.googleapis.com. That is because this change will break some of the logic in the metric_descriptor_cache.go. However, this change is only meant to be used for Google internal endpoints. A warning is given above the flag definition that this is the case.

### DIFF
--- a/prometheus-to-sd/main.go
+++ b/prometheus-to-sd/main.go
@@ -59,7 +59,9 @@ var (
 	omitComponentName = flag.Bool("omit-component-name", true,
 		"If metric name starts with the component name then this substring is removed to keep metric name shorter.")
 	debugPort = flag.Uint("port", 6061, "Port on which debug information is exposed.")
-
+	// Warning: Does not work for custom.googleapis.com endpoints.
+	gaugeToCumulativeWhitelist = flag.String("gauge-to-cumulative-whitelist", "",
+	        "Comma-separated list of Prometheus Gauge metrics which will be converted to Stackdriver Cumulatives.")
 	customMetricsPrefix = "custom.googleapis.com"
 )
 
@@ -70,7 +72,7 @@ func main() {
 	defer glog.Flush()
 	flag.Parse()
 
-	sourceConfigs := config.SourceConfigsFromFlags(source, component, host, port, whitelisted)
+	sourceConfigs := config.SourceConfigsFromFlags(source, component, host, port, whitelisted, gaugeToCumulativeWhitelist)
 
 	gceConf, err := config.GetGceConfig(*metricsPrefix)
 	podConfig := &config.PodConfig{
@@ -113,10 +115,18 @@ func main() {
 
 func readAndPushDataToStackdriver(stackdriverService *v3.Service, gceConf *config.GceConfig, podConfig *config.PodConfig, sourceConfig config.SourceConfig) {
 	glog.Infof("Running prometheus-to-sd, monitored target is %s %v:%v", sourceConfig.Component, sourceConfig.Host, sourceConfig.Port)
+
+	// Parse whitelist for metrics that will undergo Prometheus Gauge to Stackdriver Cumulative translation.
+        var gaugeToCumulativeWhitelistedList []string
+	if gaugeToCumulatativeWhitelist != "" {
+		gaugeToCumulativeWhitelistedList = strings.Split(gaugeToCumulativeWhitelist, ",")
+	}
+
 	commonConfig := &config.CommonConfig{
 		GceConfig:     gceConf,
 		PodConfig:     podConfig,
 		ComponentName: sourceConfig.Component,
+		GaugetoCumulativeWhitelist: gaugeToCumulativeWhitelistedList,
 	}
 	metricDescriptorCache := translator.NewMetricDescriptorCache(stackdriverService, commonConfig, sourceConfig.Component)
 	signal := time.After(0)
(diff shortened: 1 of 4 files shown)


prometheus-to-sd/main.go
```diff
@@ -70,7 +72,7 @@ func main() {
 	defer glog.Flush()
 	flag.Parse()
 
-	sourceConfigs := config.SourceConfigsFromFlags(source, component, host, port, whitelisted)
+	sourceConfigs := config.SourceConfigsFromFlags(source, component, host, port, whitelisted, gaugeToCumulativeWhitelist)
```
loburm: This thing is not going to compile and you don't need it.

prometheus-to-sd/main.go
```diff
@@ -153,7 +163,7 @@ func readAndPushDataToStackdriver(stackdriverService *v3.Service, gceConf *confi
 		if strings.HasPrefix(commonConfig.GceConfig.MetricsPrefix, customMetricsPrefix) {
 			metricDescriptorCache.UpdateMetricDescriptors(metrics, sourceConfig.Whitelisted)
 		}
-		ts := translator.TranslatePrometheusToStackdriver(commonConfig, sourceConfig.Whitelisted, metrics, metricDescriptorCache)
+		ts := translator.TranslatePrometheusToStackdriver(commonConfig, sourceConfig.Whitelisted, gaugeToCumulativeWhitelistedList, metrics, metricDescriptorCache)
```
loburm: The same, it doesn't compile and we you don't need (it's a part of common config)

prometheus-to-sd/translator/translator.go
```diff
@@ -143,6 +143,11 @@ func translateOne(config *config.CommonConfig,
 		EndTime: time.Now().UTC().Format(time.RFC3339),
 	}
 	metricKind := extractMetricKind(mType)
+	// Prometheus Cumulative -> Stackdriver Gauge translation
+	if metricKind == "GAUGE" && isMetricGaugeToCumulativeWhitelisted(name, config) {
+		metricKind = "CUMULATIVE"
+		interval.StartTime = start.UTC().Format(time.RFC3339)
```
loburm: This is not needed, it's going to be processed by branch below.

prometheus-to-sd/translator/translator.go
```diff
@@ -321,3 +326,13 @@ func getResourceLabels(config *config.CommonConfig) map[string]string {
 		"container_name": "",
 	}
 }
+
+// Checks if a metric is whitelisted for Gauge -> Cumulative translation.
```
loburm: Comment should start with a function name.

prometheus-to-sd/translator/translator.go
```diff
@@ -143,6 +143,11 @@ func translateOne(config *config.CommonConfig,
 		EndTime: time.Now().UTC().Format(time.RFC3339),
 	}
 	metricKind := extractMetricKind(mType)
+	// Prometheus Cumulative -> Stackdriver Gauge translation
+	if metricKind == "GAUGE" && isMetricGaugeToCumulativeWhitelisted(name, config) {
+		metricKind = "CUMULATIVE"
+		interval.StartTime = start.UTC().Format(time.RFC3339)
+	}
```
loburm: You need also to modify setValue function, because it's still going to use wrong field.

prometheus-to-sd/translator/translator_test.go
```diff
@@ -316,9 +342,12 @@ func TestTranslatePrometheusToStackdriver(t *testing.T) {
...
 	for metricName, metric := range metrics {
-		metricDescriptor := MetricFamilyToMetricDescriptor(commonConfig, metric, getOriginalDescriptor(metricName))
-		expectedMetricDescriptor := metricDescriptors[metricName]
-		assert.Equal(t, metricDescriptor, expectedMetricDescriptor)
+		// Don't perform this test for the gaugeToCumulativeMetric since it will break it.
+		if metricName != gaugeToCumulativeMetric {
+			metricDescriptor := MetricFamilyToMetricDescriptor(commonConfig, metric, getOriginalDescriptor(metricName))
+			expectedMetricDescriptor := metricDescriptors[metricName]
+			assert.Equal(t, metricDescriptor, expectedMetricDescriptor)
+	        }
```
loburm: Remove one tab.

prometheus-to-sd/main.go
rramkumar1: Sorry, somehow that stuck around when I was cleaning things up

prometheus-to-sd/main.go
rramkumar1: Done

prometheus-to-sd/translator/translator.go
rramkumar1: Done

prometheus-to-sd/translator/translator_test.go
rramkumar1: I didn't see any extra tab on my end. Does it look okay now?

prometheus-to-sd/translator/translator.go
rramkumar1: Even though I change metricKind, the Prometheus type does not change. For example, if metricKind is changed from "GAUGE" to "CUMULATIVE", we still want to get the value out of the Prometheus Gauge since the variable "metric" passed into setValue would still be Gauge

loburm:
Closing this PR, as we have agreed that this should be fixed on the side of metric source.

PR #73, opened 2017-11-28
Stop using incorrect requestheader CA

MaciekPytel:
Fixes: https://github.com/kubernetes/kubernetes/issues/55805

After this change I can get HPA with custom metrics working on GKE, 
without any RBAC-related issues.


DirectXMan12:
For future reference: if you don't specify requestheader client CA, the generic API server code will fetch it from  a special configmap created by the main Kube API server, assuming that's configured correctly.  This also works for the main client CA file.

kawych:
LGTM

PR #83, opened 2017-12-22
Accept context in translator.SendToStackdriver

jkohen:
I'd like to reuse this method in branch prometheus_writer for integration with Prometheus server, which expects to pass a deadline through the context. context.TODO() should be a no-op: https://golang.org/pkg/context/#TODO


jkohen:
I ended up copying and modifying way more than this, so no longer relevant.

PR #91, opened 2018-01-16
remove get from kubectl logs

dzolotusky:
I think that logs is a kubectl command and not a resource that you load with the get command.


x13n:
Good catch, thanks!

PR #94, opened 2018-02-05
Initial version of fluent-gcp-scaler

x13n:


fluentd-gcp-scaler/scaler.sh
```diff
@@ -0,0 +1,101 @@
...
+do
+  reset_to_defaults
+  apply_scaling
+  build_flags
+  if [ ${REQUESTS} ] || [ ${LIMITS} ]
+  then
+    echo "Running: kubectl set resources -n ${NAMESPACE} ds ${DS_NAME} ${REQUESTS} ${LIMITS}"
+    kubectl set resources -n ${NAMESPACE} ds ${DS_NAME} ${REQUESTS} ${LIMITS}
+  else
+    echo "No requests nor limits provided, doing nothing and going back to sleep."
```
kawych: nit: I'd prefer to remove this line and add timestamp to the log line where the scaler is actually doing something. This will make logs a little clearer.

fluentd-gcp-scaler/scaler.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/sh
```
kawych: Can you add an example deployment? And maybe specify what env variables are required?

fluentd-gcp-scaler/scaler.sh
```diff
@@ -0,0 +1,101 @@
...
+}
+
+build_flags() {
+  add_request cpu ${CPU_REQUEST}
+  add_request memory ${MEMORY_REQUEST}
+  add_limit cpu ${CPU_LIMIT}
+  add_limit memory ${MEMORY_LIMIT}
+}
+
+if [ -z ${DS_NAME} ]; then
```
kawych: Can you make DS_NAME and SCALING_POLICY flags? I'm fine with adding a TODO if it's just an initial version.

fluentd-gcp-scaler/scaler.sh
```diff
@@ -0,0 +1,101 @@
...
+  MEMORY_REQUEST=${DEFAULT_MEMORY_REQUEST}
+  CPU_LIMIT=${DEFAULT_CPU_LIMIT}
+  MEMORY_LIMIT=${DEFAULT_MEMORY_LIMIT}
+}
+
+apply_scaling() {
+  if ! kubectl get scalingpolicies -n ${NAMESPACE} ${SCALING_POLICY}
+  then
+    echo "${SCALING_POLICY} not found in namespace ${NAMESPACE}, using defaults."
+    exit 0
```
kawych: Is your intention to exit or return here? The message suggests that the script should continue running and reset resource requirements to defaults.

fluentd-gcp-scaler/scaler.sh
```diff
@@ -0,0 +1,101 @@
...
+  REQUESTS=
+  LIMITS=
+  CPU_REQUEST=${DEFAULT_CPU_REQUEST}
+  MEMORY_REQUEST=${DEFAULT_MEMORY_REQUEST}
+  CPU_LIMIT=${DEFAULT_CPU_LIMIT}
+  MEMORY_LIMIT=${DEFAULT_MEMORY_LIMIT}
+}
+
+apply_scaling() {
+  if ! kubectl get scalingpolicies -n ${NAMESPACE} ${SCALING_POLICY}
```
kawych: Since ScalingPolicy is not a regular k8s resource, please add a comment and link to the CRD definition.

fluentd-gcp-scaler/scaler.sh
```diff
@@ -0,0 +1,101 @@
...
+  then
+    echo "${SCALING_POLICY} not found in namespace ${NAMESPACE}, using defaults."
+    exit 0
+  fi
+  for resource_class in request limit
+  do
+    for resource_type in cpu memory
+    do
+      TMP=$(kubectl get scalingpolicies -n ${NAMESPACE} ${SCALING_POLICY} \
+        -o=jsonpath="{.spec.containers[?(@.name=='fluentd-gcp')].resources.${resource_class}s[?(@.resource=='${resource_type}')].base}")
```
kawych: Can you pass a DS name via flag/env var?

fluentd-gcp-scaler/scaler.sh
x13n: Good catch, should be return, thanks!

x13n:
I don't want this to be generic. This is a short term hack for a specific addon. Long term, we should have (cluster-proportional?) autoscaling. By not making this generic I'm increasing the chances this will not be adopted by other addons. If someone will fork this code and reuse it for something else, this is fine, but I'm willing to maintain this only for fluentd-gcp.

fluentd-gcp-scaler/scaler.sh
x13n: Ack, done.

fluentd-gcp-scaler/scaler.sh
x13n: Added.

fluentd-gcp-scaler/scaler.sh
x13n: DS name is passed by a env var. 'fluentd-gcp' here is a container name and I don't see a reason to set it externally.

fluentd-gcp-scaler/scaler.sh
x13n: Added.

fluentd-gcp-scaler/scaler.sh
x13n: Added.

fluentd-gcp-scaler/scaler.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/sh
+
+SLEEP_SECONDS=${SLEEP_SECONDS:-60}
+NAMESPACE=${NAMESPACE:-kube-system}
+DEFAULT_CPU_REQUEST=${CPU_REQUEST}
```
kawych: Is it intended to be "" if not specified? Would it be more explicit to write DEFAULT_CPU_REQUEST=${CPU_REQUEST:-}

x13n: Done.

PR #95, opened 2018-02-05
Update fluentd-gcp to use the 0.3 base image

tallclair:
This should use the latest debian-base image to get the latest CVE fixes.


x13n:
Should we immediately release a new fluentd-gcp version or are those fixes non-critical?

tallclair:
non-critical. We should make sure a new version goes out before 1.10 though.

x13n:
Ack, lgtm.

PR #101, opened 2018-02-14
Remove hack to copy systemd libraries from host in fluentd-gcp container.

filbranden:
Remove hack to copy systemd libraries from host in fluentd-gcp container. That hack was present to make fluentd-gcp be able to access journal files from COS, which are created using LZ4 compression. But debian-base:0.3 (on which this is now based) already supports both +XZ and +LZ4 compression, so this hack is no longer needed.

That hack breaks on, say, CentOS 7, where the systemd libraries link to other versions of libraries that are not available in the fluentd-gcp.

For more context, see some discussion [here](https://github.com/kubernetes/kubernetes/pull/41407#issuecomment-363971712).

Also clean up the /run.sh shell script and effectively stop using it altogether (though still ship it, for compatibility.) See individual commits for the individual cleanup steps.

Bump version to 2.0.16 to incorporate these changes.

cc @crassirostris who introduced the /host/lib hack to this container.
cc @adityakali to help test this in COS.
cc @x13n who pushed latest changes to this container. (Can you help us build official versions with these commits?)
cc @tallclair who bumped debian-base from 0.1 to 0.3 in this container.

### DIFF
--- a/fluentd-gcp-image/run.sh
+++ b/fluentd-gcp-image/run.sh
@@ -20,10 +20,4 @@
 # For systems without journald
 mkdir -p /var/log/journal
 
-# Copy host libsystemd into image to avoid compatibility issues.
-if [ ! -z "$(ls /host/lib/libsystemd* 2>/dev/null)" ]; then
-  rm /lib/x86_64-linux-gnu/libsystemd*
-  cp -a /host/lib/libsystemd* /lib/x86_64-linux-gnu/
-fi
-
-/usr/local/bin/fluentd $@
+exec /usr/local/bin/fluentd "$@"
--- a/fluentd-gcp-image/Makefile
+++ b/fluentd-gcp-image/Makefile
@@ -26,7 +26,7 @@
 .PHONY:	build push
 
 PREFIX=gcr.io/google-containers
-TAG = 2.0.15
+TAG = 2.0.16
 
 build:
 	docker build --pull -t $(PREFIX)/fluentd-gcp:$(TAG) .


fluentd-gcp-image/run.sh
```diff
@@ -17,13 +17,4 @@
...
-mkdir -p /var/log/journal
-
-# Copy host libsystemd into image to avoid compatibility issues.
-if [ ! -z "$(ls /host/lib/libsystemd* 2>/dev/null)" ]; then
-  rm /lib/x86_64-linux-gnu/libsystemd*
-  cp -a /host/lib/libsystemd* /lib/x86_64-linux-gnu/
-fi
-
-/usr/local/bin/fluentd $@
+exec /usr/local/bin/fluentd "$@"
```
tallclair: Can we delete this script now?

filbranden: As I pointed out in the PR/commit description, I suggest keeping it around just in case someone is still running it with something like:

```shell
docker run gcr.io/google-containers/fluentd-gcp:latest /run.sh $FLUENTD_ARGS
```

I don't think it's that detrimental to keep it around, since the space it takes is really minimal...

So I think I'd rather not delete it, or at least not delete it right away. I could be swayed the other way.

tallclair: Oh sorry, I missed that.  I suggest adding this to the script, so we can clean it up later (and it's clear why it's still here):

```shell
echo >&2 "DEPRECATED: use /usr/local/bin/fluentd directly"
```

filbranden:
> I suggest adding this to the script, so we can clean it up later ...

Done! PTAL

crassirostris:
@tallclair thanks a lot for reviewing!

@filbranden could you please squash commits?

filbranden:
> @filbranden could you please squash commits?

@crassirostris I actually intended to keep them as separate commits, since they make sense on their own.

If you really prefer them squashed, you can do so at merge time with the "Squash and merge" option in the drop down next to the merge button on GitHub (see [here](https://stackoverflow.com/a/36377634) for details and a screenshot.)

fluentd-gcp-image/Dockerfile
```diff
@@ -45,6 +45,9 @@ RUN BUILD_DEPS="make gcc g++ libc6-dev ruby-dev libffi-dev" \
               /var/log/* \
               /var/tmp/*
 
+# For systems without journald.
+RUN mkdir -p /var/log/journal
```
crassirostris: Oh, that actually won't work, because `/var/log` is mounted in runtime and this directory is dropped

filbranden:
I just saw a small snag here... In creation of /var/log/journal (in /run.sh or Dockerfile)...

Because the way this container is used, the host's /var/log is mounted into the container's /var/log, so the mkdir from /run.sh actually creates that directory in the *host's* /var/log if it doesn't exist there...

Should we remove some intermediate commits to remove that change (and related ones) from that PR?

In particular, that means that /run.sh continues to be used (so we're dropping 3 commits, the mkdir, the change to CMD and the deprecation warning added...)

@crassirostris Let me know what you'd like to do about this one!

fluentd-gcp-image/Dockerfile
filbranden: Yes I seem to have noticed the same thing :-)

I'll rework that and send a single squashed commit. Hang on.

crassirostris:
@filbranden What you proposed (dropping some parts of the changes) sounds reasonable, thanks!

filbranden:
Done. And squashed. PTAL.

crassirostris:
LGTM, thanks!

filbranden:
Thanks @crassirostris !

Do you have plans to release a new 2.0.16 version to GCR? Let me know when that's available, then I'll push changes to Kubernetes to start using it and stop trying to mount the host's lib directory that's no longer used...

Cheers!
Filipe

x13n:
I pushed 2.0.16 to gcr.io and will modify https://github.com/kubernetes/kubernetes/pull/59808 to use it in Kubernetes. Currently this PR bumps the version to 2.0.15 which is broken anyway.

PR #102, opened 2018-02-16
Remove version tag from the GCR reference

tomoe:
Otherwise we'd have to update README at every release.

### DIFF
--- a/prometheus-to-sd/README.md
+++ b/prometheus-to-sd/README.md
@@ -7,8 +7,7 @@ k8s cluster should run on GCE or GKE.
 
 ## Container Image
 
-The latest container image can be found at
-`gcr.io/google-containers/prometheus-to-sd:v0.2.1`.
+Look at the following link: https://gcr.io/google-containers/prometheus-to-sd and pick the latest image. 
 
 # Usage
 


x13n:
Can we just keep the latest tag in one place? The easiest solution seems to be to just write in the readme to look into Makefile and read the TAG var.

crassirostris:
Well, just removing a tag makes it confusing: we don't export the `latest` tag, so trying to download such an image will fail

Maybe we should phrase it ``Look at the following link: `https://gcr.io/google-containers/prometheus-to-sd` and pick the latest image``

tomoe:
> Maybe we should phrase it Look at the following link: `https://gcr.io/google-containers/prometheus-to-sd` and pick the latest image

+1
Also, why not adding ":latest" tag when the newest release is pushed?

tomoe:
Updated the working with @crassirostris 's suggestion.

x13n:
Thanks. Can you also make this link a link? Right now one has to copy&paste it.

tomoe:
@x13n Done!

x13n:
Thanks!

PR #106, opened 2018-02-21
Add staging deployment for Custom Metrics - Stackdriver Adapter.

kawych:


custom-metrics-stackdriver-adapter/adapter-staging.yaml
```diff
@@ -0,0 +1,136 @@
+apiVersion: v1
```
x13n: Please add at the top of this file:
- What is it for (testing?)
- When not to use it (production?)
- What to use instead of this, if necessary (something from k8s repo?)

kawych: Done

PR #107, opened 2018-02-21
Add GCB config for fluentd-gcp build

rvkubiak:
Adds a Google Container Builder config to build the fluentd-gcp image.
This will be used for build automation of Kubernetes addon images.

### DIFF
--- a/fluentd-gcp-image/cloudbuild.yaml
+++ b/fluentd-gcp-image/cloudbuild.yaml
@@ -1,0 +1,30 @@
+timeout: 10800s
+
+substitutions:
+  { "_PREFIX": "gcr.io/k8s-image-staging",
+    "_TAG": "2.0.16" }
+
+steps:
+- name: gcr.io/cloud-builders/git
+  id: git-clone
+  entrypoint: bash
+  args:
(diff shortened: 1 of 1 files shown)


x13n:
I think fluentd-gcp-image going away in favor of stackdriver-logging-agent. cc @crassirostris @bmoyles0117

rvkubiak:
@tallclair Tim, do we want to remove fluentd-gcp-image from the list of core addons?

rvkubiak:
As long as the images are in the addons directory included in the Kubernetes release, on any supported release branch (1.7, 1.8, 1.9), then we need to support them (and fix vulnerabilities). There might be a transition period where it's not worth the effort to automate some images that are going away.

Even if this image is being replaced in the future, we still want to automate the builds for current and previously supported releases.

x13n:
Ack, makes sense. Can you also add some doc/information on how this config will be used?

FWIW, the PR to replace this image is already in flight: https://github.com/kubernetes/kubernetes/pull/59128

There are other addons in this repository, are you going to add GCB configs for them as well?

rvkubiak:

PR #109, opened 2018-02-23
Create cluster roles for HPA for Custom Metrics - Stackdriver Adapter staging deployment

kawych:


x13n:
Closing, since this essentially does the same thing as #110 and the other PR already has some discussion.

PR #110, opened 2018-02-27
Add RBAC for HPA Controller to read external metrics

bskiba:
Add a role for external metrics reader.
Bind HPA Controller to that role.

Without this, HPA controller does not have permission to read external metrics api.

### DIFF
--- a/custom-metrics-stackdriver-adapter/adapter-staging.yaml
+++ b/custom-metrics-stackdriver-adapter/adapter-staging.yaml
@@ -138,3 +138,30 @@ spec:
     name: custom-metrics-stackdriver-adapter
     namespace: custom-metrics
   version: v1beta1
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: external-metrics-reader
+rules:
+- apiGroups:
+  - "external.metrics.k8s.io"
+  resources:
+  - "*"
+  verbs:
+  - list
+  - get
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: external-metrics-reader
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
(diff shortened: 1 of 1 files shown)


bskiba:
@danielmk can you review?

custom-metrics-stackdriver-adapter/adapter-staging.yaml
```diff
@@ -138,3 +138,30 @@ spec:
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: external-metrics-reader
+rules:
+- apiGroups:
+  - "external.metrics.k8s.io"
+  resources:
+  - "*"
```
x13n: Does it have to read everything? Is it possible to have external metrics on e.g. custom resources? Also, this will enable reading secrets, is it necessary?

bskiba: Can we go forward with this for now since this is the staging version? I see the alpha adapter version does the same thing and the staging version should only be used for testing at the moment.

x13n: Well, ok, but please leave a TODO here to have at least a slight chance of fixing this in the future.

kawych: It has to read all types of resources. In current version of proposal: https://github.com/kubernetes/community/pull/1802, the resource corresponds to metric name, so HPA needs access to all of them. In earlier version of the proposal, resource name was "externalmetrics", but I assume that the implementation of new proposal for HPA is already merged and this won't change. @MaciekPytel to confirm.

kawych:
LGTM
Note that it is almost exact duplicate of already existing PR: https://github.com/GoogleCloudPlatform/k8s-stackdriver/pull/109. Was that intentional due to "watch" permission?

bskiba:
No, I simply didn't see the other PR, sorry.

bskiba:
I guess this is obsolete now.

PR #112, opened 2018-03-02
Add Google Container Builder config to build kubelet-to-gcm image

rvkubiak:
Creates a GCB config that will be used for build automation of
Kubernetes addon images.


rvkubiak:

PR #114, opened 2018-03-02
Add Google Container Builder config for event-exporter

rvkubiak:
Part of future plans to automate the builds for all core Kubernetes addons. Provides build provenance by using Google Container Builder to ensure builds are created from checked in source and logs are maintained. Pushes to the k8s-image-staging registry which will be used for promoting images to eliminate the need for humans to push directly to google-containers.


rvkubiak:

PR #115, opened 2018-03-06
Fall back to GCE zone if cluster-location is not available from Metadata Server

kawych:


custom-metrics-stackdriver-adapter/pkg/config/gce_config.go
```diff
@@ -44,7 +45,11 @@ func GetGceConfig(metricsPrefix string) (*GceConfig, error) {
 
 	location, err := gce.InstanceAttributeValue("cluster-location")
 	if err != nil {
-		return nil, fmt.Errorf("error while getting cluster location: %v", err)
+		glog.Warningf("Failed to retrieve cluster location, Falling back to local zone: %s", err)
```
x13n: One nit:
s/Falling/falling/

kawych: Done, also fixed a small bug with trailing white characters.

x13n:
Thanks, lgtm.

PR #117, opened 2018-03-09
 Implement translation between External Metrics API and Stackdriver API

kawych:
This PR adds a code for translation between External Metrics API and Stackdriver API. To use this code, Custom Metrics - Stackdriver Adapter will have to expose External Metrics API endpoint, similarly to this PR: https://github.com/kubernetes-incubator/custom-metrics-apiserver/pull/15


kawych:
@MaciekPytel PTAL

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -201,6 +203,30 @@ func (p *StackdriverProvider) ListAllMetrics() []provider.MetricInfo {
...
+	if err != nil {
+		return nil, err
+	}
+	return &external_metrics.ExternalMetricValueList{
+		Items: externalMetricItems,
+	}, nil
+}
+
+func unescapeMetricName(metricName string) string {
+	return strings.Replace(metricName, "|\\", "/", -1)
```
MaciekPytel: Can we change this to just a pipe ('|')? I wanted a longer escape sequence when it was meant for multiple different systems, but AFAIK pipe is not allowed in sd names, so it's good enough by itself.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
```diff
@@ -137,6 +161,44 @@ func (t *Translator) GetRespForMultipleObjects(response *stackdriver.ListTimeSer
...
 
+// GetRespForExternalMetric translates Stackdriver response to list of External Metrics
+func (t *Translator) GetRespForExternalMetric(response *stackdriver.ListTimeSeriesResponse, metricName string) ([]external_metrics.ExternalMetricValue, error) {
+	if len(response.TimeSeries) < 1 {
+		return nil, provider.NewExternalMetricNotFoundError(metricName)
+	}
+	metricValues := []external_metrics.ExternalMetricValue{}
+	// Find time series with specified labels matching
+	// Stackdriver API doesn't allow complex label filtering (i.e. "label1 = x AND (label2 = y OR label2 = z)"),
+	// therefore only part of the filters is passed and remaining filtering is done here.
```
MaciekPytel: This comments says there should be some filtering here, but I see nothing of the sort. Am I missing something? Or is there some code missing here?

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
```diff
@@ -187,6 +249,29 @@ func getNodeNames(list *v1.NodeList) []string {
...
+	return false
+}
+
+func splitMetricLabel(labelName string) (string, string, error) {
+	for _, prefix := range allowedLabelPrefixes {
+		if strings.HasPrefix(labelName, prefix+".") {
+			return prefix, strings.TrimPrefix(labelName, prefix+"."), nil
+		}
+	}
+	return "", "", fmt.Errorf("TODO")
```
MaciekPytel: ??

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
```diff
@@ -265,6 +358,87 @@ func (t *Translator) legacyFilterForPods(podIDs []string) string {
 	return fmt.Sprintf("resource.label.pod_id = one_of(%s)", strings.Join(podIDs, ","))
 }
 
+func (t *Translator) filterForSelector(metricSelector labels.Selector) (string, error) {
+	requirements, selectable := metricSelector.Requirements()
+	if !selectable {
+		return "", provider.NewOperationNotSupportedError("TODO")
```
MaciekPytel: Add error message / error handling

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
```diff
@@ -265,6 +358,87 @@ func (t *Translator) legacyFilterForPods(podIDs []string) string {
...
+func (t *Translator) getMetricLabels(series *stackdriver.TimeSeries) map[string]string {
+	metricLabels := map[string]string{}
+	for label, value := range series.Metric.Labels {
+		metricLabels["metric.label."+label] = value
+	}
+	metricLabels["resource.type"] = series.Resource.Type
+	for label, value := range series.Resource.Labels {
+		metricLabels["resource.label."+label] = value
+	}
+	return metricLabels
```
MaciekPytel: Should we include metadata labels here? I don't know if we should, but we do allow them in query.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator_test.go
```diff
@@ -369,6 +372,97 @@ func TestTranslator_GetMetricsFromSDDescriptorsResp(t *testing.T) {
...
+	request, err := translator.GetExternalMetricRequest("custom.googleapis.com/my/metric/name", labels.NewSelector().Add(*req1, *req2, *req3, *req4))
+	if err != nil {
+		t.Fatalf("Translation error: %s", err)
+	}
+	expectedRequest := sdService.Projects.TimeSeries.List("projects/my-project").
+		Filter("metric.type = \"custom.googleapis.com/my/metric/name\" " +
+			"AND metric.label.my_label > 86 " +
+			"AND NOT resource.label.namespace_name = one_of(\"default\",\"kube-system\") " +
+			"AND resource.label : pod_name " +
+			"AND resource.type = \"k8s_pod\"").
```
MaciekPytel: Is the order of conditions well defined or does the ordering in expected query just randomly work?

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator_test.go
```diff
@@ -369,6 +372,97 @@ func TestTranslator_GetMetricsFromSDDescriptorsResp(t *testing.T) {
...
+		IntervalStartTime("2017-01-02T13:00:00Z").
+		IntervalEndTime("2017-01-02T13:01:00Z").
+		AggregationPerSeriesAligner("ALIGN_NEXT_OLDER").
+		AggregationAlignmentPeriod("60s")
+	if !reflect.DeepEqual(*request, *expectedRequest) {
+		t.Errorf("Unexpected result. Expected \n%s,\n received: \n%s", *expectedRequest, *request)
+	}
+}
+
+func TestTranslator_GetExternalMetricRequest(t *testing.T) {
```
MaciekPytel: It feels like the test name is missing suffix

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator_test.go
kawych: They're sorted by key: https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/selector.go#L100

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
kawych: I think we should not. Metric and resource labels are exported with each metric. Metadata labels are exported by different channel to provide more information about given resource to Stackdriver. But resource labels include a complete set of information required to identify the resource.

For example, pods are identified by resource labels: cluster name, location, project id, namespace and name. Metadata contain information such as node name, pod labels, etc.

MaciekPytel:
I think we could use a few more UT covering more filters (ex. a combination of few filters where only one is incorrect, or a query with more than one condition of given type (label1 = value1 and label2 = value2)).

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator_test.go
MaciekPytel: Cool, that answers my question.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
kawych: Done

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
kawych: Done

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator_test.go
kawych: Done

kawych:
Added a test for more filters where one is incorrect. I've added the other case to one of existing ones, although I think it doesn't do much, since all requirements are translated separately.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/translator.go
MaciekPytel: So you can use metadata labels to query based on pod labels, and every metric value will come with a set of labels identifying it (namespace + name), correct? Makes sense.

PR #124, opened 2018-03-15
Allow for custom metric endpoints

### LINKED ISSUE #108: non standard url for prom metrics
Is there a way to pull prom metrics from http://localhost/xyx instead of always defaulting to  http://localhost/metrics ?

When I specify 
```
 command:
        - /monitor
        - --source=:http://localhost:80/promMetrics
        - --pod-id=$(POD_ID)
        - --namespace-id=$(POD_NAMESPACE)


```
It still defaults to /metrics

deejay1:
Spring 5 by default the actuator endpoint /status/prometheus,
so we should allow for setting it in config.
The default /metrics is left alone. Fixes #108.


deejay1:
Fixed conflicts

deejay1:
Any hope on this? We're near one month now from the initial commit and two months since #108 was opened ;)

deejay1:
@loburm Added a defaultMetricsPath const and used it throughout the test file also. Unfortunately the build is broken probably because https://github.com/golang/lint/issues/396 :(

loburm:
I have rerun by myself, everything looks good.

Thanks once again.

PR #125, opened 2018-03-15
Bugfix: don't call kubectl set resources when there would be no effect.

x13n:
Step towards fixing https://github.com/kubernetes/kubernetes/issues/61190, which is k8s 1.10.0 release blocker.

### DIFF
--- a/fluentd-gcp-scaler/scaler.sh
+++ b/fluentd-gcp-scaler/scaler.sh
@@ -8,8 +8,8 @@ DEFAULT_CPU_LIMIT=${CPU_LIMIT:-}
 DEFAULT_MEMORY_LIMIT=${MEMORY_LIMIT:-}
 
 reset_to_defaults() {
-  REQUESTS=
-  LIMITS=
+  REQUESTS_FLAG=
+  LIMITS_FLAG=
   CPU_REQUEST=${DEFAULT_CPU_REQUEST}
   MEMORY_REQUEST=${DEFAULT_MEMORY_REQUEST}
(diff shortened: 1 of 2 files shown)


fluentd-gcp-scaler/scaler.sh
```diff
@@ -80,6 +80,39 @@ build_flags() {
...
+  if needs_update requests.cpu ${CPU_REQUEST}; then NEED_UPDATE=true; fi
+  if needs_update requests.memory ${MEMORY_REQUEST}; then NEED_UPDATE=true; fi
+  if needs_update limits.cpu ${CPU_LIMIT}; then NEED_UPDATE=true; fi
+  if needs_update limits.memory ${MEMORY_LIMIT}; then NEED_UPDATE=true; fi
+  if ! ${NEED_UPDATE}
+  then
+    log "Nothing to update."
+    return
+  fi
+  if [ ${REQUESTS_FLAG} ] || [ ${LIMITS_FLAG} ]
```
kawych: Is it possible that code reaches this point and $REQUESTS_FLAG and/or $LIMIT_FLAG are empty?

x13n: Yes, but only under quite bizarre circumstances - when all values are not set anywhere (neither in defaults, nor in scaling policy), but resources are configured in the daemonset. The code above will figure it should update something, but both flags will be empty. Shouldn't happen in practice.

x13n:
If there will be no further comments, I will merge it in 1h.

PR #126, opened 2018-03-15
Update fluentd-gcp and event-exporter base image

crassirostris:
Use the new version of base images. Additionally, update the names to the new ones

@cjcullen Please review


x13n:
Is this intended to be cherrypicked into older k8s releases? After https://github.com/kubernetes/kubernetes/pull/59128 our fluentd-gcp image is no longer used.

crassirostris:
@x13n Yes

PR #127, opened 2018-03-15
 Fetch changes from template custom-metrics-apiserver to expose External Metrics API

kawych:
This PR introduces to Custom Metrics - Stackdriver Adapter the same changes as https://github.com/kubernetes-incubator/custom-metrics-apiserver/pull/15. The other PR is in review, to avoid being blocked on it, this PR introduces the same changes as its current shape. Any future changes to this PR will be introduced here in follow-up PRs (if applicable).


custom-metrics-stackdriver-adapter/pkg/apiserver/apiserver.go
```diff
@@ -40,7 +41,8 @@ var (
 )
 
 func init() {
-	install.Install(groupFactoryRegistry, registry, Scheme)
+	cminstall.Install(groupFactoryRegistry, registry, Scheme)
+	eminstall.Install(groupFactoryRegistry, registry, Scheme)
```
MaciekPytel: Shouldn't we only install api group if it's enabled?

custom-metrics-stackdriver-adapter/pkg/apiserver/installer/emhandlers.go
```diff
@@ -0,0 +1,143 @@
...
+	reqScope.Namer = MetricsNaming{
+		handlers.ContextBasedNaming{
+			GetContext:         ctxFn,
+			SelfLinker:         a.group.Linker,
+			ClusterScoped:      false,
+			SelfLinkPathPrefix: gpath.Join(a.prefix, scope.ParamName()) + "/",
+		},
+	}
+
+	externalMetricHandler := metrics.InstrumentRouteFunc("LIST", "external-metrics", "", "", restfulListResource(lister, nil, reqScope, false, a.minRequestTimeout))
```
MaciekPytel: "external-metrics" is no longer part of path in the final version of API. Should it be used here?

custom-metrics-stackdriver-adapter/pkg/apiserver/apiserver.go
kawych: Do you mean to move everything from the init() function to completedConfig.New()? I can do that, but on the other hand, this only registers types and api group, so it doesn't have any effect visible for caller.

custom-metrics-stackdriver-adapter/pkg/apiserver/installer/emhandlers.go
kawych: I think it's just an arbitrary name that doesn't correspond to anything in the path, similar to https://github.com/kubernetes-incubator/custom-metrics-apiserver/blob/master/pkg/apiserver/installer/installer.go#L297.

PR #128, opened 2018-03-15
Update example to work with old and new Stackdriver resource model

kawych:


bskiba:
Code LGTM, but we need to explain what's going on in the readme. For example there is a part of troubleshooting that says, make sure you are setting pod id correctly.

kawych:
As discussed offline, the documentation will be updated when new resource model is available publicly.

PR #138, opened 2018-03-28
Use instance_name instead of hostname for metrics.

loburm:


loburm:
I would avoid overcomplicating this PR with such unit test. It will require us to introduce one additional abstraction level, because gce metadata package doesn't provide any interface.

Not sure if such tests will provide us any additional value.

x13n:
Ack, other than that, lgtm

PR #140, opened 2018-03-28
Support DELTA and CUMULATIVE metrics

kawych:


kawych:
@MaciekPytel PTAL

MaciekPytel:
I would be tempted to add some caching of metric types (without it we do 2x more requests), but caching is hard and it may well be a premature optimization. Otherwise great stuff.

/lgtm

kawych:
Fixed one more bug. I didn't implement cache for exactly the reason that it is probably premature, agreed that some cache would help.
Adding one more reviewer from logmon.

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -95,7 +95,11 @@ func (p *StackdriverProvider) GetRootScopedMetricByName(groupResource schema.Gro
 	if err != nil {
 		return nil, err
 	}
-	stackdriverRequest, err := p.translator.GetSDReqForNodes(&v1.NodeList{Items: []v1.Node{*matchingNode}}, metricName)
+	metricKind, err := p.translator.GetMetricKind(p.translator.config.MetricsPrefix + "/" + metricName)
```
x13n: Can you do that metric name building in GetMetricKind argument? It would be way more natural if it was prepending the prefix itself.

kawych: Done

PR #148, opened 2018-04-20
Bumped fluentd from 0.12.32 to 0.14.25

yliaog:
0.14.25 is stable, tested, and has the feature to support nanosecond timestamp.


yliaog:
@deejay1 please take another look

yliaog:
@x13n @loburm could you this a look at this PR?

evanj:
The big issue that I think this will fix is the fact that Google Kubernetes Engine truncates logs to second resolution. I think the underlying cause is this fluent-plugin-google-cloud bug, which was fixed in v0.6.9 (already included in this image I think): https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/131

However, that will only work with fluentd >= 0.14 since before that it represents time as seconds: https://www.fluentd.org/blog/fluentd-v0.14.0-has-been-released

evanj:
See the GKE issue for this also: https://issuetracker.google.com/issues/35903605

yliaog:
@loburm it seems that fluentd 1.2.5 is stable, and qualified, updated this PR to upgrade to v1.2.5. PTAL

fluentd-gcp-image/Gemfile
```diff
@@ -1,8 +1,8 @@
 source 'https://rubygems.org'
 
-gem 'fluentd', '~>0.12.32'
+gem 'fluentd', '~>1.2.5'
```
igorpeshansky: Note: we have only qualified this version of fluentd as it pertains to the Docker image we distribute (i.e., the one built from https://github.com/GoogleCloudPlatform/google-fluentd/blob/master/docker/Dockerfile). There is no evidence that it will work properly for this Docker image. Should this not be obsoleted now that we are providing one?
@qingling128 @bmoyles0117 FYI.

PR #149, opened 2018-04-24
version bump for event exporter

cezarygerard:


kawych:
@cezarygerard @loburm 
What's the convention for naming event exporter releases? In general, I think we increase the last digit for the bugfixes. Since this is a new feature, shouldn't it be 0.2.0?

loburm:
AFAIK there is no strong rule, because in this repo we are not performing usual patches. But I agree that in current case it makes sense to increase it to 0.2.0.

PR #154, opened 2018-05-09
[Custom Metrics] Retrieve metrics from last 5 minutes, but compute rate only from 1 minute period

kawych:


kawych:
/cc @MaciekPytel @aleksandra-malinowska 
PTAL

Issue #161, opened 2018-06-05
[kubelet-to-gcm] Dropping metrics for containers with same name and start time


serathius:
Looks like mechanism for deduplicating data points (https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/e474fade9b2195967fb84cae4cd64b1613bfa202/kubelet-to-gcm/monitor/kubelet/translate.go#L187)
is not handling cases where we have two pods with containers that have same name and start time. 
Same start dates are common when pods are deployed using as kubelet static pods. 

/cc @loburm

PR #162, opened 2018-06-05
Move container stats deduplication logic under pod

serathius:
Deduplicating for whole scrape caused dropping metrics when there are
multiple pods with containers that have same name.
closes https://github.com/GoogleCloudPlatform/k8s-stackdriver/issues/161

### DIFF
--- a/kubelet-to-gcm/monitor/kubelet/translate_test.go
+++ b/kubelet-to-gcm/monitor/kubelet/translate_test.go
@@ -21,6 +21,9 @@ import (
 	"testing"
 	"time"
 
+	fuzz "github.com/google/gofuzz"
+
+	"k8s.io/kubernetes/pkg/api/unversioned"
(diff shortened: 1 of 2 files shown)


kubelet-to-gcm/monitor/kubelet/translate_test.go
```diff
@@ -196,3 +199,117 @@ func TestTranslator(t *testing.T) {
...
+	return stats.PodStats{
+		PodRef:      stats.PodReference{Name: "test-pod", Namespace: "test-namespace", UID: "UID_test-pod"},
+		StartTime:   unversioned.NewTime(time.Now()),
+		Containers:  containers,
+		Network:     getNetworkStats(),
+		VolumeStats: []stats.VolumeStats{*getVolumeStats()},
+	}
+}
+
+func getFsStats() *stats.FsStats {
```
loburm: This function is not used.

serathius: done

PR #166, opened 2018-06-12
[Custom Metrics] Remove priority field from apiservice definitions

kawych:
This caused Custom Metrics - Stackdriver Adapter e2e tests to fail.


x13n:
I have no context here. Why these fields were causing problems?

kawych:
The priority field was removed from API definition: https://github.com/kubernetes/kube-aggregator/blob/master/pkg/apis/apiregistration/types.go, therefore it causes validator errors in k8s 1.11

PR #167, opened 2018-06-15
[Custom Metrics] Return empty list instead of 404 when the metric is found but there are no timeseries

kawych:


MaciekPytel:
I don't know the code well enough to figure out exactly how the output will change. Based on a quick glance on HPA code I think an empty list will generally result in an error being reported by HPA. Basically the end result will be the same as before (HPA creating event and setting status condition), but the error message will be different.

MaciekPytel:
@kawych Can you describe in more detail what happened before this change and what will happen after it?

kawych:
I've seen customers confused about error messages for metrics that exist but without timeseries for provided selector, so I want to make this clearer. Also, when customers call External Metrics API endpoint directly, they get 404 and I think the more accurate outcome will be an empty list if the metric is correct.

For Custom Metrics API query about metrics for multiple objects I've implemented the same change, but after a second thought I think that the number of returned metrics should be exactly equal to the number of matched objects (e.g. pods), so the 404 error would make more sense here as well. I'm going to revert the change to Custom Metrics API in this PR.

MaciekPytel:
Agreed. I just don't understand how this PR changes it to empty list? It seems to return NewMetricNotFoundForError if the metric list is empty. Will that result in returning 200 with an empty list result?

kawych:
@MaciekPytel 
I'm merging this, but feel free to post any other questions here, I'll address them as a follow up.

PR #172, opened 2018-06-21
Vary configuration based on the existence of GCE metadata.

bmoyles0117:


kawych:
Looks good, adding @loburm to take another look as prom-to-sd expert.

bmoyles0117:
You're absolutely right, thanks Marian!

PR #174, opened 2018-07-13
Replace base image of event-exporter with image without security vulnerabilities.

loburm:


kawych:
/lgtm

Can you say in the description it's event exporter base image?

loburm:
@kawych done, thanks for the review

PR #177, opened 2018-07-24
[prometheus-to-sd] Allow using prometheus labels for SD resource

### LINKED ISSUE #173: [Feature Request] Ability to override resource labels with prometheus labels
I am working on adding a [cAdvisor daemonset](https://github.com/google/cadvisor/tree/master/deploy/kubernetes) with an [example integration with the prometheus-to-sd sidecar](https://github.com/google/cadvisor/blob/master/deploy/kubernetes/overlays/examples/stackdriver-sidecar.yaml)

Prometheus-to-sd currently just uses the [namespace_id, pod_id, and container_name provided by the config](https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/prometheus-to-sd/translator/translator.go#L316).  However, for cAdvisor, or other monitoring daemonsets that publish metrics about other pods/containers in the cluster, adding namespace_id, pod_id, and container_name as prometheus labels results in stackdriver labels pointing to the "correct" container, but the monitored resource points to the cAdvisor container (as that is where the metrics originate from).

I propose either allowing prometheus labels to override the monitored resource attributes by default (if provided), or adding the ability to configure which prometheus label is used for each monitored resource attribute.  E.g. `--container-name-label`, `--pod-id-label`, etc, and allowing only these, or the `--pod-id`, `--namespace-id` flags to be set.  The latter has the advantage of being able to perform "remapping" of labels from the target pod to monitored resource attributes in stackdriver, and means we don't need to match labels exactly (e.g. I could specify `--namespace-id-label=namespace_name` if that was how the producer labeled metrics).

dashpole:
Fixes #173.  See the issue for details.

```release-note
Adds the `container-name-label`, `pod-id-label`, `namespace-id-label` flags which allow prometheus labels to be used for SD resource labels.
```

### DIFF
--- a/event-exporter/Makefile
+++ b/event-exporter/Makefile
@@ -19,7 +19,7 @@ BINARY_NAME = event-exporter
 
 PREFIX = gcr.io/google-containers
 IMAGE_NAME = event-exporter
-TAG = v0.2.1
+TAG = v0.2.2
 
 build:
 	${ENVVAR} godep go build -a -o ${BINARY_NAME}


dashpole:
not sure why this closed.  reopened in #178

PR #178, opened 2018-07-24
[prometheus-to-sd] Allow using prometheus labels for SD resource

### LINKED ISSUE #173: [Feature Request] Ability to override resource labels with prometheus labels

dashpole:
Fixes #173.  See the issue for details.

```release-note
Adds the `container-name-label`, `pod-id-label`, `namespace-id-label` flags which allow prometheus labels to be used for SD resource labels.
```

### DIFF
--- a/prometheus-to-sd/README.md
+++ b/prometheus-to-sd/README.md
@@ -55,3 +55,18 @@ Example of [deployment](https://github.com/GoogleCloudPlatform/k8s-stackdriver/b
 used to monitor
 [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) component, that is used to collect
 different metrics about the state of k8s cluster.
+
+## Container monitoring agents
+
+Container monitoring agents, such as [cAdvisor](https://github.com/google/cadvisor#cadvisor), 
+collect metrics about containers other than itself.  Monitoring agents use prometheus
+labels for the namespace, pod, and container to differentiate metrics from different
+containers.  To use these prometheus labels as the monitored resource labels in
+stackdriver, include `namespaceIdLabel`, `podIdLabel`, and `containerNameLabel` in the 
+`source` flag in this format: 
+`component-name:http://host:port?whitelisted=a,b,c&namespaceIdLabel=d&podIdLabel=e&containerNameLabel=f`.
+Note that prometheus labels used for the monitored resource are not included as labels in stackdriver.
+
+For example, if prom-to-sd scraped the prometheus metric: 
+`container_cpu_usage_seconds{d="my-namespace",e="abc123",f="my-app",g="production"} 1.02030405e+09`
+It would be displayed in stackdriver with the namespace, `my-namespace`, the pod id, `abc123`, and the container name, `my-app`.  The only stackdriver label for this metric would be `g="production"`.
--- a/prometheus-to-sd/config/pod_config_test.go
+++ b/prometheus-to-sd/config/pod_config_test.go
@@ -1,0 +1,210 @@
+/*
+Copyright 2018 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"testing"
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+func TestIsMetricLabel(t *testing.T) {
+	for _, tc := range []struct {
+		desc   string
+		config *podConfigImpl
+		label  string
+		want   bool
+	}{
+		{
+			desc:   "empty",
+			config: &podConfigImpl{},
+			label:  "foo",
+			want:   true,
+		},
+		{
+			desc: "containerNameLabel matches",
+			config: &podConfigImpl{
+				containerNameLabel: "foo",
+				podIdLabel:         "bar",
+				namespaceIdLabel:   "abc",
+			},
+			label: "foo",
+			want:  false,
+		},
+		{
+			desc: "podIdLabel matches",
+			config: &podConfigImpl{
+				containerNameLabel: "foo",
+				podIdLabel:         "bar",
+				namespaceIdLabel:   "abc",
+			},
+			label: "bar",
+			want:  false,
+		},
+		{
+			desc: "namespaceIdLabel matches",
+			config: &podConfigImpl{
+				containerNameLabel: "foo",
+				podIdLabel:         "bar",
+				namespaceIdLabel:   "abc",
+			},
+			label: "abc",
+			want:  false,
+		},
+		{
+			desc: "none match",
+			config: &podConfigImpl{
+				containerNameLabel: "foo",
+				podIdLabel:         "bar",
+				namespaceIdLabel:   "abc",
+			},
+			label: "xyz",
+			want:  true,
+		},
+	} {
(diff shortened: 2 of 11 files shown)


dashpole:
cc @x13n @loburm

prometheus-to-sd/config/pod_config.go
```diff
@@ -16,16 +16,61 @@ limitations under the License.
...
+	podIdLabel         string
+	namespaceIdLabel   string
+	containerNameLabel string
+}
+
+func (p *podConfigImpl) IsMetricLabel(labelName string) bool {
+	return labelName != p.podIdLabel && labelName != p.containerNameLabel && labelName != p.namespaceIdLabel
+}
+
+func (p *podConfigImpl) GetPodInfo(labels []*dto.LabelPair) (containerName, podId, namespaceId string) {
```
loburm: Please, add tests for the new logic.

prometheus-to-sd/main.go
```diff
@@ -57,6 +57,12 @@ var (
 		"Name of the pod in which monitored component is running.")
 	namespaceId = flag.String("namespace-id", "",
 		"Namespace name of the pod in which monitored component is running.")
+	containerNameLabel = flag.String("container-name-label", "",
```
loburm: Also please update documentation.

prometheus-to-sd/config/pod_config.go
dashpole: done

prometheus-to-sd/main.go
dashpole: udpated.

dashpole:
I addressed comments.  Let me know if there is anything else needed.

x13n:
I'm not sure if adding flags that are inherently pod-specific make sense when we're about to deploy prom-to-sd as a DaemonSet.

/cc @kawych @Szetty

kawych:
My understanding is that these flags are not necessarily pod-specific, but can be. I'm not fully aware of context of this change, if the prometheus labels with podId, namespace and container name are specific for pods, I'd recommend to add these label names per-source, similarly to the list of whitelisted metrics.

For example: `--source=<component_name>:http://localhost:1?whitelisted=qps&podIdLabel=pod-id&namespaceIdLabel=namespace-id`

dashpole:
I added the fields per-source.  How soon will we transition to a daemonset for prom-to-sd?

prometheus-to-sd/config/pod_config_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+	dto "github.com/prometheus/client_model/go"
+)
+
+func TestIsMetricLabel(t *testing.T) {
+	for _, tc := range []struct {
+		desc     string
+		config   *podConfigImpl
+		label    string
+		expected bool
```
loburm: want instead of expected

The same for other tests.

prometheus-to-sd/config/pod_config_test.go
```diff
@@ -0,0 +1,210 @@
...
+				namespaceIdLabel:   "abc",
+			},
+			label:    "xyz",
+			expected: true,
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			actual := tc.config.IsMetricLabel(tc.label)
+			if actual != tc.expected {
+				t.Fatalf("Unexpected result; got %v, expected %v", actual, tc.expected)
```
loburm: want instead of expected

prometheus-to-sd/main.go
```diff
@@ -60,6 +60,12 @@ var (
 		"Namespace name of the pod in which monitored component is running.")
 	zoneOverride = flag.String("zone-override", "",
 		"Name of the zone to override the default one (in which component is running).")
+	containerNameLabel = flag.String("container-name-label", "",
```
loburm: Considering that you moved this to per source configuration, we don't need those flags anymore. Also don't forget to revert known-flags.txt file.

prometheus-to-sd/README.md
```diff
@@ -13,7 +13,7 @@ Look at the following link: https://gcr.io/google-containers/prometheus-to-sd an
 
 For scraping metrics from the component it's name, host, port and metrics should passed
 through the flag `source` in the next format:
-`component-name:http://host:port?whitelisted=a,b,c`. If whitelisted part is
+`component-name:http://host:port?whitelisted=a,b,c&podIdLabel=d&namespaceIdLabel=e&containerNameLabel=f`. If whitelisted part is
```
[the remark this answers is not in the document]
dashpole: done.

prometheus-to-sd/config/pod_config_test.go
dashpole: done

prometheus-to-sd/config/pod_config_test.go
dashpole: done

prometheus-to-sd/main.go
dashpole: done

prometheus-to-sd/config/pod_config_test.go
```diff
@@ -0,0 +1,210 @@
...
+		},
+		{
+			Name:  &namespaceLabel,
+			Value: &namespace,
+		},
+	}
+	for _, tc := range []struct {
+		desc                  string
+		config                *podConfigImpl
+		expectedContainerName string
```
loburm: want here and in other places

prometheus-to-sd/config/source_config.go
```diff
@@ -96,10 +90,7 @@ func (config *SourceConfig) UpdateWhitelistedMetrics(list []string) {
...
 // SourceConfigsFromFlags creates a slice of SourceConfig's base on the provided flags.
 func SourceConfigsFromFlags(source flags.Uris, component *string, host *string, port *uint, whitelisted *string, podId *string, namespaceId *string) []SourceConfig {
-	podConfig := PodConfig{
-		PodId:       *podId,
-		NamespaceId: *namespaceId,
-	}
+	podConfig := NewPodConfig(*podId, *namespaceId, "", "", "")
 	var sourceConfigs []SourceConfig
 	for _, c := range source {
 		if sourceConfig, err := parseSourceConfig(c, podConfig); err != nil {
```
loburm: You also need to extract all values in parseSourceConfig function above, otherwise you will not be able to configure it using --source. 

dynamic_source actually is going to be removed.

prometheus-to-sd/config/pod_config_test.go
dashpole: done

prometheus-to-sd/config/source_config.go
dashpole: done

x13n:
/lgtm

@loburm, is there anything else you'd like to change in this PR or can we merge it?

PR #179, opened 2018-07-25
Change package manager from the deprecated Godeps to dep

Szetty:
Remove Godeps related files
Add dep related files
Change the build command


prometheus-to-sd/Gopkg.toml
```diff
@@ -0,0 +1,38 @@
+# Gopkg.toml example
```
serathius: Do we need example?

prometheus-to-sd/Makefile
```diff
@@ -16,13 +16,13 @@ all: build
 
 ENVVAR = GOOS=linux GOARCH=amd64 CGO_ENABLED=0
 PREFIX = gcr.io/google-containers
-TAG = v0.2.6
+TAG = v0.2.7
```
kawych: This line looks like a mistake, you shouldn't bump prom-to-sd version in this PR

serathius: We don't need to bump version for changes in tooling.

PR #180, opened 2018-07-25
Change docker build

Szetty:
Build the project using multi-stage docker building in the docker container


prometheus-to-sd/Dockerfile
```diff
@@ -12,10 +12,20 @@
...
+COPY . ./
+RUN CGO_ENABLED=0 GOOS=linux go build -o /usr/bin/monitor
 
+FROM gcr.io/google-containers/debian-base-amd64:0.3
 RUN clean-install ca-certificates
-
-#   nobody:nobody
 USER 65534:65534
-COPY monitor /
+COPY --from=builder /usr/bin/monitor /usr/bin/
```
serathius: Please don't change binary location

prometheus-to-sd/Dockerfile
```diff
@@ -12,10 +12,20 @@
...
 
-FROM gcr.io/google-containers/debian-base-amd64:0.3
+FROM golang:1.9-alpine as builder
+ARG DEP_VERSION="0.4.1"
+RUN apk update && apk add bash git
+ADD https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 /usr/bin/dep
+RUN chmod +x /usr/bin/dep
+WORKDIR ${GOPATH}/src/github.com/GoogleCloudPlatform/k8s-stackdriver/prometheus-to-sd
+COPY Gopkg.toml Gopkg.lock ./
+RUN dep ensure -vendor-only
```
serathius: I don't like idea of using dep when building image. Can we just use preexisting vendor?

Szetty: it is used only to verify dependencies, I think it is good to have one more step of check

prometheus-to-sd/Dockerfile
Szetty: why?

serathius: Because we are depending on it in kubernetes manifests. It's a breaking change that is not needed.

prometheus-to-sd/Dockerfile
serathius: verifying dependencies could be part of test pipeline not building image

prometheus-to-sd/Dockerfile
```diff
@@ -12,10 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/google-containers/debian-base-amd64:0.3
+FROM golang:1.9-alpine as builder
```
Szetty: I am using 1.9 as Godeps was using this version for building the code before

serathius:
LGTM

loburm:
lgtm

PR #182, opened 2018-07-26
[prometheus-to-sd] Add PodConfig in SourceConfig

Szetty:
Until now all the sources were on the same pod, but as we prepare for changing prometheus to a daemon set, now it makes sense to include the pod configuration in the source configuration
This is a small refactoring for the upcoming daemon set changes

### DIFF
--- a/prometheus-to-sd/config/source_config.go
+++ b/prometheus-to-sd/config/source_config.go
@@ -34,12 +34,19 @@ type SourceConfig struct {
 	Port        uint
 	Path        string
 	Whitelisted []string
+	PodConfig   PodConfig
+}
+
+// PodConfig describes the pod in which the monitored component is running.
+type PodConfig struct {
+	PodId       string
+	NamespaceId string
 }
 
 const defaultMetricsPath = "/metrics"
 
 // newSourceConfig creates a new SourceConfig based on string representation of fields.
-func newSourceConfig(component string, host string, port string, path string, whitelisted string) (*SourceConfig, error) {
+func newSourceConfig(component string, host string, port string, path string, whitelisted string, podConfig PodConfig) (*SourceConfig, error) {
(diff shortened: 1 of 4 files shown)


prometheus-to-sd/config/source_config.go
```diff
@@ -34,12 +34,19 @@ type SourceConfig struct {
 	Port        uint
 	Path        string
 	Whitelisted []string
+	PodConfig   PodConfig
+}
+
+// PodConfig describes pod in which current component is running.
```
serathius: Outdated comment

prometheus-to-sd/config/source_config.go
```diff
@@ -34,12 +34,19 @@ type SourceConfig struct {
...
+type PodConfig struct {
```
serathius: To you need to move PodConfig definition?

prometheus-to-sd/main.go
```diff
@@ -73,13 +73,13 @@ func main() {
...
 
-	sourceConfigs := config.SourceConfigsFromFlags(source, component, host, port, whitelisted)
-
-	gceConf, err := config.GetGceConfig(*metricsPrefix, *zoneOverride)
-	podConfig := &config.PodConfig{
+	podConfig := config.PodConfig{
 		PodId:       *podId,
 		NamespaceId: *namespaceId,
 	}
+	sourceConfigs := config.SourceConfigsFromFlags(source, component, host, port, whitelisted, podConfig)
```
kawych: nit: podConfig should be created inside this function from flags as well.

kawych:
LGTM apart from one nit.

prometheus-to-sd/main.go
Szetty: resolved

PR #183, opened 2018-07-27
[prometheus-to-sd] Add daemon set mode to prometheus-to-sd

Szetty:
Add flags for daemon set mode
Take source config from Kubernetes API in daemon set mode


serathius:
Please split changes to vendor directory to separate PR

prometheus-to-sd/config/api_server_client.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, err
+	}
+	pods := podResponse.Items
+	if len(pods) < 0 {
+		return nil, errors.New("no pods for monitoring were found")
+	}
+	return getConfigsFromPods(pods)
+}
+
+func setupKubernetesPodsApi(namespace string) (corev1.PodInterface, error) {
```
serathius: It will be more reusable to have function that returns k8s client, not particular api.

prometheus-to-sd/config/api_server_client.go
```diff
@@ -0,0 +1,68 @@
...
+		podConfig := PodConfig{
+			PodId:       pod.Name,
+			NamespaceId: pod.Namespace,
+		}
+		port := pod.Annotations["prom-to-sd.googleapis.com/port"]
+		whitelisted := pod.Annotations["prom-to-sd.googleapis.com/whitelisted"]
+		sourceConfig, err := newSourceConfig(pod.Name, ip, port, defaultMetricsPath, whitelisted, podConfig)
+		if err != nil {
+			message := fmt.Sprintf("error with pod %s: %v", pod.Name, err)
+			return nil, errors.New(message)
```
serathius: I don't think that missing pods to monitor should result in failure

prometheus-to-sd/config/api_server_client.go
```diff
@@ -0,0 +1,68 @@
...
+
+func getConfigsFromPods(pods []core.Pod) ([]SourceConfig, error) {
+	var sourceConfigs []SourceConfig
+	for _, pod := range pods {
+		ip := pod.Status.PodIP
+		podConfig := PodConfig{
+			PodId:       pod.Name,
+			NamespaceId: pod.Namespace,
+		}
+		port := pod.Annotations["prom-to-sd.googleapis.com/port"]
```
serathius: Can you move names of annotations to consts

prometheus-to-sd/main.go
```diff
@@ -111,6 +114,23 @@ func main() {
 	<-make(chan int)
 }
 
+func getSourceConfigs(gceConfig *config.GceConfig) (sourceConfigs []config.SourceConfig) {
```
serathius: don't use names return for simple case

prometheus-to-sd/main.go
```diff
@@ -111,6 +114,23 @@ func main() {
...
+	if *asDaemonSet {
```
serathius: If asDaemonSet flag disables functionality of normal sources, you should add validation that forbids passing both

prometheus-to-sd/main.go
```diff
@@ -111,6 +114,23 @@ func main() {
...
+		glog.Info("Taking source configs from api-server")
+		var err error
+		sourceConfigs, err = config.GetConfigsFromApiServer(gceConfig, *namespace)
```
serathius: Pods are dynamic, what happens if some pod is added/removed. You should redesign whole pipeline to allow dynamic sources

PR #185, opened 2018-08-02
[prometheus-to-sd] Fix small bugs, add path in dynamic source

Szetty:
Add path to dynamic source configs
Handle empty path
Fix component name bug
Add tests for empty path


Szetty:
cc @loburm

PR #186, opened 2018-08-02
[prometheus-to-sd] Improve source validation

Szetty:
Add tests for source validation

### DIFF
--- a/prometheus-to-sd/config/dynamic_source.go
+++ b/prometheus-to-sd/config/dynamic_source.go
@@ -47,6 +47,9 @@ func validateSources(sources flags.Uris) (map[string]url.URL, error) {
 		if source.Key == "" {
 			return nil, errors.New("component name should NOT be empty for any dynamic source")
 		}
+		if source.Val.Port() == "" {
+			return nil, errors.New("port should NOT be empty for any dynamic source")
+		}
 		sourceMap[source.Key] = source.Val
 	}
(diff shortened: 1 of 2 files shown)


Szetty:
cc @loburm

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -47,6 +47,9 @@ func validateSources(sources flags.Uris) (map[string]url.URL, error) {
 		if source.Key == "" {
 			return nil, errors.New("component name should NOT be empty for any dynamic source")
 		}
+		if source.Val.Port() == "" {
+			return nil, errors.New("port should NOT be empty for any dynamic source")
```
loburm: Why do we need a second check?

https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/prometheus-to-sd/config/source_config.go#L50

Szetty: to have the validation of all the required fields in one place, we should not always rely on the fact that the port will be validated after, also it is good to fail fast

Szetty: also the port seems to be validated for static resources before too
https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/985653a490cab36c496aa666f410a0696816ee61/prometheus-to-sd/config/source_config.go#L79

PR #188, opened 2018-08-07
[prometheus-to-sd] Implement watch for dynamic sources

Szetty:
Watch the components specified by the component name filters
Translate watch event to source config change
Implement goroutine that listens to source config changes
Add watch dependency in lock file

### DIFF
--- a/prometheus-to-sd/config/dynamic_source.go
+++ b/prometheus-to-sd/config/dynamic_source.go
@@ -9,17 +9,36 @@ import (
 	"github.com/golang/glog"
 	core "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
 const (
-	podNamespace = "kube-system"
-	nameLabel    = "k8s-app"
+	podNamespace       = "kube-system"
+	componentNameLabel = "k8s-app"
 )
 
-// SourceConfigsFromDynamicSources takes pod specifications from the Kubernetes API and maps them to source configs.
-func SourceConfigsFromDynamicSources(gceConfig *GceConfig, sources []flags.Uri) ([]SourceConfig, error) {
+// ChangeType defines the possible types of events.
+type ChangeType string
+
+// Possible change types
+const (
+	Added   ChangeType = "ADDED"
+	Deleted ChangeType = "DELETED"
+	Unknown ChangeType = "UNKNOWN"
+)
+
+// SourceConfigChange describes how one of the dynamic sources have changed.
+// Is generated when watch generates a new event.
+type SourceConfigChange struct {
+	ChangeType   ChangeType
+	SourceConfig SourceConfig
+}
+
+// CreateDynamicSourcesStream takes source specifications and creates a watch for them.
+// Every time a new event comes, it is mapped to source configs and sent through the channel returned by this function
+func CreateDynamicSourcesStream(gceConfig *GceConfig, sources []flags.Uri) (<-chan SourceConfigChange, error) {
 	if len(sources) == 0 {
 		return nil, nil
 	}
@@ -31,11 +50,8 @@ func SourceConfigsFromDynamicSources(gceConfig *GceConfig, sources []flags.Uri) 
 	if err != nil {
 		return nil, err
 	}
-	podResponse, err := kubeApi.CoreV1().Pods(podNamespace).List(createOptionsForPodSelection(gceConfig.Instance, sourceMap))
-	if err != nil {
-		return nil, err
-	}
-	return getConfigsFromPods(podResponse.Items, sourceMap), nil
+	podOptions := createOptionsForPodSelection(gceConfig.Instance, sourceMap)
+	return getSourceConfigChangeChannel(kubeApi, sourceMap, podOptions)
 }
 
 func validateSources(sources flags.Uris) (map[string]url.URL, error) {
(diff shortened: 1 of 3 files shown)


Szetty:
cc @kawych @serathius

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -18,8 +19,28 @@ const (
...
+)
+
+// SourceConfigChange describes how one of the dynamic sources have changed.
+// Is generated when watch generates a new event.
+type SourceConfigChange struct {
+	ChangeType   ChangeType
+	SourceConfig SourceConfig
+	EventObject  string
+	Err          error
+}
```
serathius: I think introducing internal events is unnecessary complication.

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -31,11 +52,13 @@ func SourceConfigsFromDynamicSources(gceConfig *GceConfig, sources []flags.Uri)
 	if err != nil {
 		return nil, err
 	}
-	podResponse, err := kubeApi.CoreV1().Pods(podNamespace).List(createOptionsForPodSelection(gceConfig.Instance, sourceMap))
+	podApi := kubeApi.CoreV1().Pods(podNamespace)
```
serathius: this variable can be inlined

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -31,11 +52,13 @@ func SourceConfigsFromDynamicSources(gceConfig *GceConfig, sources []flags.Uri)
...
+	podOptions := createOptionsForPodSelection(gceConfig.Instance, sourceMap)
+	watchInterface, err := podApi.Watch(podOptions)
```
serathius: I think creating watch interface should be moved to `getSourceConfigChangeChannel` function

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -18,8 +19,28 @@ const (
...
+// Is generated when watch generates a new event.
+type SourceConfigChange struct {
+	ChangeType   ChangeType
+	SourceConfig SourceConfig
+	EventObject  string
+	Err          error
+}
+
+// GetChannelForDynamicSources takes pod specifications from the Kubernetes API and maps them to source configs.
+func GetChannelForDynamicSources(gceConfig *GceConfig, sources []flags.Uri) (<-chan SourceConfigChange, error) {
```
serathius: I don't think adding Channel to names is good change in naming.

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -72,32 +95,62 @@ func createOptionsForPodSelection(nodeName string, sources map[string]url.URL) v
...
 
 func mapToSourceConfig(componentName string, url url.URL, ip string, podConfig PodConfig) (*SourceConfig, error) {
 	port := url.Port()
 	whitelisted := url.Query().Get("whitelisted")
 	return newSourceConfig(componentName, ip, port, url.Path, whitelisted, podConfig)
 }
+
+func getSourceConfigChangeChannel(watchInterface watch.Interface, sources map[string]url.URL) <-chan SourceConfigChange {
+	sourceConfigChannel := make(chan SourceConfigChange)
+	eventChannel := watchInterface.ResultChan()
```
serathius: I think watching will send you updates, but not current state. Can you check that?

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -72,32 +95,62 @@ func createOptionsForPodSelection(nodeName string, sources map[string]url.URL) v
...
+	pod, ok := event.Object.(*core.Pod)
+	if !ok {
+		sourceConfigChange.ChangeType = Error
+		sourceConfigChange.Err = errors.New("not a pod object")
+		return sourceConfigChange
+	}
+	sourceConfig, err := getConfigFromPod(*pod, sources)
+	if err != nil {
+		sourceConfigChange.ChangeType = Error
+		sourceConfigChange.Err = err
```
serathius: Do you need to create error events and send them instead of handling them locally?

prometheus-to-sd/config/dynamic_source.go
Szetty: I think this is the correct way to go, I need to send SourceConfig and at least a type, so we need some sort of structure, also handling events in the main is not that correct, and it would depend too much on the dynamic source internal things

prometheus-to-sd/config/dynamic_source.go
Szetty: no, we can handle them locally too

prometheus-to-sd/config/dynamic_source.go
Szetty: It will also send the current state of all watched resources when you first start to watch

prometheus-to-sd/config/dynamic_source.go
Szetty: done

prometheus-to-sd/config/dynamic_source.go
Szetty: done

prometheus-to-sd/config/dynamic_source.go
Szetty: done

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -31,11 +50,8 @@ func SourceConfigsFromDynamicSources(gceConfig *GceConfig, sources []flags.Uri)
 	if err != nil {
 		return nil, err
 	}
-	podResponse, err := kubeApi.CoreV1().Pods(podNamespace).List(createOptionsForPodSelection(gceConfig.Instance, sourceMap))
```
kawych: Please correct me if I missed some part of your code, but how do you get an initial list of pods? Do you get from watch the events that happened before prom-to-sd creation?

Szetty: This is what we discovered with Marian last week, and discussed with Marek about yesterday. When you start the watch, you will get an ADDED event for all the pods that are already started and working, so this way you do not need to query for the initial pod list

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -72,32 +88,65 @@ func createOptionsForPodSelection(nodeName string, sources map[string]url.URL) v
...
+				glog.Errorf("%v", err)
+			} else {
+				sourceConfigChannel <- *sourceConfigChange
+			}
+		}
+	}()
+	return sourceConfigChannel, nil
+}
+
+func mapPodWatchToSourceConfigChange(event watch.Event, sources map[string]url.URL) (*SourceConfigChange, error) {
```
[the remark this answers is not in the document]
Szetty: this is a better name yes

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -18,8 +19,26 @@ const (
 	nameLabel    = "k8s-app"
```
kawych: nit: This name is a little ambiguous, maybe "componentNameLabel"?

prometheus-to-sd/main.go
```diff
@@ -138,40 +135,64 @@ func readAndPushDataToStackdriver(stackdriverService *v3.Service, gceConf *confi
...
-			metricDescriptorCache.ValidateMetricDescriptors(metrics, sourceConfig.Whitelisted)
+	}
+}
+
+func handleDynamicConfigurations(stackdriverService *v3.Service, gceConf *config.GceConfig, sourceConfigChangeCh <-chan config.SourceConfigChange) {
+	channels := make(map[string]chan bool)
+	for sourceConfigChange := range sourceConfigChangeCh {
+		switch sourceConfigChange.ChangeType {
+		case config.Added, config.Modified:
+			if sourceConfigChange.SourceConfig.Host != "" && channels[sourceConfigChange.SourceConfig.Component] == nil {
```
serathius: What checking not empty host is used for? are you expecting it? in what circumstances?

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -72,32 +88,65 @@ func createOptionsForPodSelection(nodeName string, sources map[string]url.URL) v
 		nameList += key + ","
 	}
 	labelSelector := fmt.Sprintf("%s in (%s)", nameLabel, nameList)
-	return v1.ListOptions{
+	listOptions := v1.ListOptions{
```
kawych: What happens if a customer creates their own pod with k8s-app label, for example "k8s-app: heapster"? You don't rely on anything that identifies pod unambiguously (like pod namespace + pod name, or just pod id), therefore you can get multiple events related to different pods, but AFAIU your code will think it's the same pod.

Note that this is a realistic scenario, for example some customers may disable stackdriver monitoring and deploy their own heapster using deployments from here: https://github.com/kubernetes/heapster/tree/master/deploy/kube-config.

Szetty: I am using pod namespace, which is kube-system, I am not sure that we can rely on pod name or pod id though

prometheus-to-sd/main.go
Szetty: We need the host/ip to be able to make requests to the component, when a pod is firstly created it does not have an ip yet

prometheus-to-sd/config/dynamic_source.go
Szetty: corrected

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -71,33 +87,81 @@ func createOptionsForPodSelection(nodeName string, sources map[string]url.URL) v
...
-			NamespaceId: pod.Namespace,
-		}
-		sourceConfig, err := mapToSourceConfig(componentName, source, pod.Status.PodIP, podConfig)
-		if err != nil {
-			glog.Warning("could not create source config for pod %s: %v", pod.Name, err)
-		}
-		sourceConfigs = append(sourceConfigs, *sourceConfig)
+func getConfigFromPod(pod core.Pod, sources map[string]url.URL) (*SourceConfig, error) {
+	if pod.Status.PodIP == "" {
+		return nil, errors.New("pod ip is missing")
```
serathius: This will be triggered for every new pod that is waiting to be scheduled. Maybe we should silence it ?

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -71,33 +87,81 @@ func createOptionsForPodSelection(nodeName string, sources map[string]url.URL) v
...
+	if !ok {
+		return nil, fmt.Errorf("not a pod object %+v", event.Object)
+	}
+	sourceConfig, err := getConfigFromPod(*pod, sources)
+	if err != nil {
+		return nil, fmt.Errorf("could not map pod %s to source config: %v", pod.Name, err)
+	}
+	sourceConfigChange := SourceConfigChange{
+		ChangeType:   getChangeType(event.Type),
+		SourceConfig: *sourceConfig,
```
serathius: What happens when you get Delete event with pod without IP?

prometheus-to-sd/main.go
```diff
@@ -138,40 +135,66 @@ func readAndPushDataToStackdriver(stackdriverService *v3.Service, gceConf *confi
...
+}
+
+func handleDynamicConfigurations(stackdriverService *v3.Service, gceConf *config.GceConfig, sourceConfigChangeCh <-chan config.SourceConfigChange) {
+	channels := make(map[string]chan bool)
+	for sourceConfigChange := range sourceConfigChangeCh {
+		sourceConfig := sourceConfigChange.SourceConfig
+		sourceConfigId := fmt.Sprintf("%s_%s_%s", sourceConfig.PodConfig.NamespaceId, sourceConfig.PodConfig.PodId, sourceConfig.Component)
+		switch sourceConfigChange.ChangeType {
+		case config.Added:
+			if channels[sourceConfigId] == nil {
```
serathius: more readable and standard to golang would be
```
if _, found :=channels[sourceConfigId]; !found {
```

prometheus-to-sd/main.go
```diff
@@ -138,40 +135,66 @@ func readAndPushDataToStackdriver(stackdriverService *v3.Service, gceConf *confi
...
+				channel := make(chan bool)
```
serathius: This channel should have size 1, so you can push message to it without blocking

prometheus-to-sd/config/dynamic_source.go
```diff
@@ -71,33 +87,81 @@ func createOptionsForPodSelection(nodeName string, sources map[string]url.URL) v
...
 	return newSourceConfig(componentName, ip, port, url.Path, whitelisted, podConfig)
 }
+
+func getSourceConfigChangeChannel(kubeApi clientset.Interface, sources map[string]url.URL, podOptions v1.ListOptions) (<-chan SourceConfigChange, error) {
+	watchInterface, err := kubeApi.CoreV1().Pods(podNamespace).Watch(podOptions)
+	if err != nil {
+		return nil, err
+	}
+	sourceConfigChannel := make(chan SourceConfigChange)
+	eventChannel := watchInterface.ResultChan()
```
serathius: Can you defer closing this channel in gorouting. Even in infinite loop is good to free resources that you take, so someone other when rewriting code will not skip it.

prometheus-to-sd/config/dynamic_source.go
Szetty: I was thinking about that, but not really seeing as possible, is it possible?

l15k4:
Guys why was this closed? I find it kind of essential for dynamic source. Otherwise the discovery happens only on start.

kawych:
The reason is, we wanted to limit prom-to-sd dependencies as an essential monitoring component, i.e. we don't want to depend on apiserver.

@loburm, I think we should either roll back all the code for dynamic discovery or revive this PR and submit it, but don't use it for monitoring system components. WDYT?

pribor:
I thought that these tools are k8s related and DaemonSet should be the recommended way how to deploy it.

l15k4:
What is wrong about using apiserver in k8s? We monitor all microservices using kamon and currently we have 17 prom-to-sd containers running in our cluster as sidecars which is a very bad design.

kawych:
The reason is, we want the monitoring to be reliable even when the apiserver is down, this is important for monitoring system components.

loburm:
Another possible issue with this approach - monitoring can fail if there is any issue with overlay network and prometheus-to-sd can't fetch metrics from that pods.

@l15k4 If you have so many pods that you want to monitor, have you considered to collect metrics using Prometheus server and then export them to Stackdriver: https://cloud.google.com/monitoring/kubernetes-engine/prometheus

PR #189, opened 2018-08-10
Allow controlling scrape and export intervals independently

### LINKED ISSUE #187: [Feature request][prom-to-sd] Decouple scrape interval from export interval
Currently prom-to-sd is exporting every batch of scraped metrics. We could decouple these two things, so that a number of scrapes would accumulate a state in memory, which then would be exported with smaller frequency. This will allow better precision of measuring metrics (e.g. when a pod dies between exports), without increasing the frequency of exports.

cc @loburm @serathius @kawych @Szetty

x13n:
Fixes #187


x13n:
> Cloning into '/home/travis/gopath/src/github.com/golang/lint'...
fatal: unable to access 'https://github.com/golang/lint/': Failed to connect to github.com port 443: Connection timed out

Travis failed to download lint package, how do I restart it?

prometheus-to-sd/main.go
```diff
@@ -63,6 +64,10 @@ var (
 		"If metric name starts with the component name then this substring is removed to keep metric name shorter.")
 	debugPort      = flag.Uint("port", 6061, "Port on which debug information is exposed.")
 	dynamicSources = flags.Uris{}
+	scrapeInterval = flag.Duration("scrape-interval", 60*time.Second,
+		"The interval between metric scrapes. If there are multiple scrapes between two exports, the last present value is exported, even when missing from last scraping.")
+	exportInterval = flag.Duration("export-interval", 60*time.Second,
+                "The interval between metric exports. Can't be less than --scrape-interval.")
```
loburm: To much whitespace.

prometheus-to-sd/translator/translator.go
```diff
@@ -42,28 +43,59 @@ var supportedMetricTypes = map[dto.MetricType]bool{
...
+func (t *timeSeriesBuilder) Build() []*v3.TimeSeries {
+	s := make([]*v3.TimeSeries, 0, len(t.timeSeries))
+	for _, v := range t.timeSeries {
+		s = append(s, v)
+	}
+	t.timeSeries = make(map[string]*v3.TimeSeries)
+	return s
+}
 
+// Update translates metrics in Prometheus format to Stackdriver format and updates the current batch.
```
loburm: Is there sense to translate metrics during every scrape? If we are going to send only 1 out 30 or 60 scrapes, then it's just a waste of CPU time. Propose to move this part of the logic to Build method and keep raw scraped metrics in timeSeriesBuilder.

prometheus-to-sd/translator/translator.go
```diff
@@ -42,28 +43,59 @@ var supportedMetricTypes = map[dto.MetricType]bool{
 
 const falseValueEpsilon = 0.001
 
-// TranslatePrometheusToStackdriver translates metrics in Prometheus format to Stackdriver format.
-func TranslatePrometheusToStackdriver(config *config.CommonConfig,
-	whitelisted []string,
-	metrics map[string]*dto.MetricFamily,
-	cache *MetricDescriptorCache) []*v3.TimeSeries {
+type timeSeriesBuilder struct {
```
loburm: I would say that is a bit unusual to return private struct, but at the same time to have all methods public. For example that means timeSeriesBuilder couldn't be passed into any function.

prometheus-to-sd/translator/translator.go
```diff
@@ -42,28 +43,59 @@ var supportedMetricTypes = map[dto.MetricType]bool{
...
+		if t.cache.IsMetricBroken(name) {
 			continue
 		}
-		t, err := translateFamily(config, metric, startTime, cache)
+		f, err := translateFamily(t.config, metric, startTime, t.cache)
 		if err != nil {
 			glog.Warningf("Error while processing metric %s: %v", name, err)
 		} else {
-			ts = append(ts, t...)
+			for _, v := range f {
```
loburm: From what I have understood you are trying to collect latest values for all possible combinations of metric labels. Not sure that it worth doing it. First it's very rare pattern when someone resets prometheus collector, so usually the last scrape will give you the same result as you are trying to achieve here.

prometheus-to-sd/main.go
kawych: nit: "Can't be lower than (...)"

prometheus-to-sd/translator/translator.go
x13n: It can be passed as an interface, if needed. Making the struct private ensures that it's created only using the New* function. See http://www.golangpatterns.info/object-oriented/constructors

prometheus-to-sd/translator/translator.go
x13n: Good point. I'll change it.

prometheus-to-sd/translator/translator.go
x13n: Ok, sounds reasonable, PTAL.

prometheus-to-sd/translator/translator.go
```diff
@@ -77,14 +102,15 @@ func OmitComponentName(metricFamilies map[string]*dto.MetricFamily, componentNam
...
-func getStartTime(metrics map[string]*dto.MetricFamily) time.Time {
+func getStartTime(metrics map[string]*metricFamilyWithTimestamp) time.Time {
 	// For cumulative metrics we need to know process start time.
 	// If the process start time is not specified, assuming it's
 	// the unix 1 second, because Stackdriver can't handle
 	// unix zero or unix negative number.
 	startTime := time.Unix(1, 0)
-	if family, found := metrics[processStartTimeMetric]; found && family.GetType() == dto.MetricType_GAUGE && len(family.GetMetric()) == 1 {
-		startSec := family.Metric[0].Gauge.Value
+	glog.Infof("getStartTime(%v)", metrics)
```
loburm: Do we need this?

x13n: Nope, leftover from debugging. Thanks for spotting!

x13n:
Thanks! I squashed commits. Once Travis check passes, I'm going to merge this.

x13n:
aaand it failed
> fatal: unable to access 'https://go.googlesource.com/tools/': Failed to connect to go.googlesource.com port 443: Connection timed out

PR #190, opened 2018-08-13
Bump prometheus-to-sd version tag for releasing

x13n:
@loburm @kawych @Szetty


x13n:
Added one more commit with doc update, PTAL.

PR #193, opened 2018-08-16
Fix metric whitelist updating in timeSeriesBuilder

x13n:


x13n:
Also, bumped the tag number.

x13n:
One more thing: I bumped base debian image.

@kawych FYI

PR #199, opened 2018-08-27
Added comparison with stackdriver-prometheus

jkohen:
The motivation is that the Stackdriver support team gets user requests to support prometheus-to-sd, even though we have an alternative that some of these users aren't aware of.

### DIFF
--- a/prometheus-to-sd/README.md
+++ b/prometheus-to-sd/README.md
@@ -70,3 +70,9 @@ Note that prometheus labels used for the monitored resource are not included as 
 For example, if prom-to-sd scraped the prometheus metric: 
 `container_cpu_usage_seconds{d="my-namespace",e="abc123",f="my-app",g="production"} 1.02030405e+09`
 It would be displayed in stackdriver with the namespace, `my-namespace`, the pod id, `abc123`, and the container name, `my-app`.  The only stackdriver label for this metric would be `g="production"`.
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


prometheus-to-sd/README.md
```diff
@@ -70,3 +70,9 @@ Note that prometheus labels used for the monitored resource are not included as
 For example, if prom-to-sd scraped the prometheus metric: 
 `container_cpu_usage_seconds{d="my-namespace",e="abc123",f="my-app",g="production"} 1.02030405e+09`
 It would be displayed in stackdriver with the namespace, `my-namespace`, the pod id, `abc123`, and the container name, `my-app`.  The only stackdriver label for this metric would be `g="production"`.
+
+## Alternatives
+
+Google develops **prometheus-to-sd** primarily for Google Kubernetes Engine to collect metrics from system services in order to support Kubernetes users. We designed the tool to be lean when deployed as a sidecar in your pod. It's intended to support only the metrics the Kubernetes team at Google needs; you can use it to monitor your applications, but the feature set is limited.
```
loburm: I'm not completely agree with sentence that it's only for Kubernetes team at Google. We have multiple cases when people are using it for monitoring their workloads and for example kube-state-metrics.

jkohen:
Marian, what text do you suggest?

prometheus-to-sd/README.md
jkohen: Marian, what text do you suggest?

jkohen: Ping, can you suggest some text? Stackdriver regularly gets customers asking for support for this tool. @piosz

PR #200, opened 2018-08-29
[kubelet-to-gcm] Split requests that are too big & use standard flag lib

x13n:


x13n:
@serathius @loburm Please review.

PR #211, opened 2018-10-04
Add support for new stackdriver kubernetes resource types guarded by a flag.

bmoyles0117:


prometheus-to-sd/translator/translator.go
```diff
@@ -351,15 +348,49 @@ func createProjectName(config *config.GceConfig) string {
 	return fmt.Sprintf("projects/%s", config.Project)
 }
 
-func getResourceLabels(config *config.CommonConfig, labels []*dto.LabelPair) map[string]string {
+func getMonitoredResourceFromLabels(config *config.CommonConfig, labels []*dto.LabelPair) *v3.MonitoredResource {
+	if config == nil || config.PodConfig == nil || config.GceConfig == nil {
```
x13n: These fields weren't nil-checked before, can this actually happen?

prometheus-to-sd/main.go
```diff
@@ -51,6 +51,8 @@ var (
 		"Namespace name of the pod in which monitored component is running.")
 	zoneOverride = flag.String("zone-override", "",
 		"Name of the zone to override the default one (in which component is running).")
+	useNewResources = flag.Bool("use-new-resources", false,
```
x13n: Please not "new" :-) Can we name it somehow? A year from now, this won't be new anymore, but the flag will stick for a while.

prometheus-to-sd/translator/translator_test.go
```diff
@@ -247,6 +247,60 @@ var metricDescriptors = map[string]*v3.MetricDescriptor{
 	},
 }
 
+func TestGetMonitoredResourceFromLabels(t *testing.T) {
```
x13n: There is a set of actual test cases here, can you t.Run() them separately?

prometheus-to-sd/main.go
bmoyles0117: Updated

bmoyles0117:
Friendly ping @x13n

prometheus-to-sd/main.go
```diff
@@ -51,6 +51,8 @@ var (
 		"Namespace name of the pod in which monitored component is running.")
 	zoneOverride = flag.String("zone-override", "",
 		"Name of the zone to override the default one (in which component is running).")
+	monitoredResourceTypes = flag.String("monitored-resource-types", "gke",
```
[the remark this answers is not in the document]
bmoyles0117: I feel like "legacy" has the same time argument as "new", I chose gke, as it references a specific part of the monitored resource itself "gke"_container, would you still prefer to use legacy?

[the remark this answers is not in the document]
x13n: Good point, although hopefully things that are "legacy" stay "legacy" while "new" becomes "currently used". How about "gke_container"? Sounds less vague and describes exactly what it is.

x13n:
Looks good to me, but Travis is still unhappy... can you make sure it passes?

l15k4:
@bmoyles0117 hey, I tried it out and there seems to be a strange error that I cannot figure out : 
```
stackdriver.go:58] Error while sending request to Stackdriver googleapi: Error 400: One or more TimeSeries could not be written: Unrecognized resource label: cluster_location: timeSeries[0-33], badRequest
```

l15k4:
@bmoyles0117 I have no idea how this can happen when I look at this https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/prometheus-to-sd/config/gce_config.go#L65-L72

bmoyles0117:
This could happen if the location provided is not a valid GCP region / zone.

l15k4:
@bmoyles0117 but how is that possible and how can I fix that? From what I can see, it is retrieved from the API in here https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/prometheus-to-sd/config/gce_config.go#L65

Could RBAC be the problem here ? Shouldn't we log these values ?

bmoyles0117:
RBAC doesn't come into play when querying the GCE Metadata Server. Can I confirm, what does your prometheus-to-sd yaml look like? Are you using a GKE-managed installation, or have you configured this yourself? If the latter, are you sure that you've set the `--monitored-resource-types` flag to "k8s"?

l15k4:
I'm running it on a k8s cluster `1.11.2-gke.18` with Stackdriver Beta enabled. The `prometheus-to-sd` container is a sidecar to `graphite-exporter` container in a pod. I also run it as a side car to many other containers that expose metrics with kamon. The configuration is : 
```
  prometheus-to-sd:
    Command:
      /monitor
      --stackdriver-prefix=custom.googleapis.com
      --monitored-resource-types=k8s
      --source=graphite-exporter:http://localhost:9095/metrics
      --pod-id=$(POD_NAME)
      --namespace-id=$(POD_NAMESPACE)
```

bmoyles0117:
@l15k4 thank you for the bug report, it was a huge oversight that I used the wrong label name for sending the cluster's location, #222 has been sent for review which should resolve this issue.

PR #218, opened 2018-11-21
Fixed resource name, updated containers to last versions available

jorgonzalez:
README.md: Fixed wording: deployment/pod.
kubernetes/prometheus-to-sd-kube-state-metrics.yaml: Updated to the last versions of the containers.


loburm:
LGTM,
Thanks

PR #219, opened 2018-12-04
Add basic metrics to kubelet-to-gcm.

loburm:
Add metrics timeseries_pushed_total and timeseries_dropped_total to measure number of timeseries pushed to Stackdriver.


loburm:
Added new metrics and squashed. PTAL.

Issue #196, opened 2018-08-17
[prometheus-to-sd] Support for SUMMARY metric types


marctc:
It seems that prometheus-to-sd is unable to translate summary metrics and throws errors every time that scrape the metrics:
```
W0817 15:04:56.690842       1 translator.go:61] Error while processing metric http_request_duration_microseconds: Metric type SUMMARY of family http_request_duration_microseconds not supported
```
However, the [stackdriver-prometheus](https://github.com/Stackdriver/stackdriver-prometheus) project actually is able to parse this kind of metric and push it to stackdriver (https://github.com/Stackdriver/stackdriver-prometheus/blob/9f5c836bfc3abea91b7473f4a57c107806099bf9/stackdriver/translate.go#L118).

Is there any reason that prometheus-to-sd does not support this kind of metrics? Are you planning to add support for summary metrics? I can work on it if you accept PRs.

chrislovecnm:
Any update on this??

chrislovecnm:
cc @GoogleCloudPlatform/logmon

erictune:
Is this not feasible or do you need help to implement this?

chrislovecnm:
What do you mean?

erictune:
I also want support for SUMMARY and I am hoping that a maintainer is listening to this issue.

jcheroske:
Running into this now myself. Curious if this project is still being maintained.

erictune:
@logicalhan Want to paste summary (e.g. from of #221) and close issue?

logicalhan:
@erictune, sure thing. 

> #221 adds partial summary metric importing ability for prom-to-sd. There are some important caveats, however:
> 
> 1. Quantile metric data is disregarded, mostly because it doesn't make a whole lot of sense to aggregate over the quantile metric data. Aggregating across this data will give you averages. You will lose quantile information.
> 2. If you want to import summary metric information, you must first ensure that the metric you are summing does not have negative values (this is despite the fact that prometheus documentation expressly allows for this possibility). Unfortunately, this must be the case due to data type impedance mismatching between prometheus and stackdriver. So if you want to use this feature, you should first ensure your metric is measuring positive values. With things like latency measurements, this shouldn't be a problem.
> 3. If you are whitelisting metrics, then you must whitelist the summary metric name + '_sum' as well as the summary metric name + "_count". This is due to the fact that this metric type maps to multiple stackdriver metrics.

/close

[...]

PR #234, opened 2019-03-01
[prometheus-to-sd] Add quantiles to summary type

AttwellBrian:
As discussed in #196, summary quantiles cannot be aggregated well. However, they are still useful and frequently used by Prometheus exporters. In order to allow stackdriver to be more of a drop in replacement for prom, this diff adds support for summary quantiles

**Test Plan**
Modified `translater_test.go`. Temporarily added the following to the `Dockerfile`. Then ran the `Dockerfile`.

```
RUN file="$(CGO_ENABLED=0 GOOS=linux go test ./translator/)"; echo $file
```


prometheus-to-sd/translator/translator_test.go
```diff
@@ -537,6 +550,17 @@ int_summary_metric_count{label="l2"} 10
...
 					},
 				},
+				{
+					Metric: &v3.Metric{
+						Labels:  map[string]string{"quantile": "0.500000",},
+						Type:   "container.googleapis.com/master/testcomponent/float_summary_metric",
+					},
+					MetricKind: "GAUGE",
+					Points: []*v3.Point{
+						// Like all gauges the translater assumes it is an integer, unfortunately :(
```
AttwellBrian: This is a seperate issue that I'll look into in a separate diff.

I've seen this cause some frustrating issues before, outside of summary metric types.

AttwellBrian:
`verify-gofmt.sh` returns no problems on my machine. So I'm not certain what to do here.

Issue #233, opened 2019-03-01
[kubelet-to-gcm] Process partial results upon errors


x13n:
kubelet-to-gcm currently fails to generate any data, even when a single container metrics are missing. We should log an error and process the rest in that case.

PR #236, opened 2019-03-01
Do not skip whole stats population for single nil stats.

Random-Liu:
Fixes https://github.com/GoogleCloudPlatform/k8s-stackdriver/issues/233.

Summary stats for CRI integration can return nil fields for terminated containers, we should not skip the whole stats population for a single nil container stats.


x13n:
Thanks for the fix!

PR #242, opened 2019-03-12
Update debian base image to 0.4.1

serathius:
New image has security fixes for 02.2019

### DIFF
--- a/fluentd-gcp-image/Dockerfile
+++ b/fluentd-gcp-image/Dockerfile
@@ -20,7 +20,7 @@
 # scope and that the Logging API has been enabled for the project
 # in the Google Developer Console.
 
-FROM gcr.io/google-containers/debian-base-amd64:0.3.2
+FROM gcr.io/google-containers/debian-base-amd64:0.4.1
 
 COPY Gemfile /Gemfile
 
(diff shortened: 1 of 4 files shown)


fluentd-gcp-image/Dockerfile
```diff
@@ -20,7 +20,7 @@
 # scope and that the Logging API has been enabled for the project
 # in the Google Developer Console.
 
-FROM gcr.io/google-containers/debian-base-amd64:0.3.2
+FROM gcr.io/google-containers/debian-base-amd64:0.4.1
```
x13n: Can we migrate this to distroless?

serathius: We need bash to run script. distroless has no version that includes bash interpreter https://github.com/GoogleContainerTools/distroless/blob/master/base/README.md

x13n: We don't need to run the script. Essentially, all that script is doing is running a binary. We might as well just run the binary...

serathius: Ok, I was thinking about "fluentd-gcp-scaler" which is implemented in bash.
I don't know enough about ruby to give definite answer, by my understanding is that ruby is interpreted language and distroless doesn't support it. I think this would be a major effort for this image which is being superseded by stackdriver agent

x13n: Ack. We should probably put a deprecation note in the README and point people to stackdriver-logging-agent instead.

serathius:
Added a commit with version bump

PR #243, opened 2019-03-14
Fluentd gcp stability fixes 2 1

serathius:
Stabilizing fluentd-gcp 2.1


fluentd-gcp-image/Makefile
```diff
@@ -18,11 +18,18 @@
...
 
 build:
 	docker build --pull -t $(PREFIX)/fluentd-gcp:$(TAG) .
 
 
 push:
 	gcloud docker -- push $(PREFIX)/fluentd-gcp:$(TAG)
+
+
+refreeze-dependencies: build
```
x13n: Why not just "update-dependencies"?

serathius: I use this work to emphasize that it will not only update Gemfile.lock but also remove version boundaries for all packages and select newest available.

x13n: This is what I would expect the word update to mean. Can you clarify in README what this command does, exactly?

serathius: I decided to rename to "update-dependencies" to match common nomenclature.

PR #244, opened 2019-03-14
 prometheus-to-sd - copies license files to the docker image

wgrzelak:


serathius:
Could you please use govendor to generate license list.

PR #248, opened 2019-03-26
prometheus-to-sd - '--source' flag can be specified multiple times

wgrzelak:


wgrzelak:
@serathius could you review?

PR #250, opened 2019-04-02
updated fluendt-gcp image

monotek:
Signed-off-by: André Bauer <andre.bauer@kiwigrid.com>

updated fluentd to 1.2.5, all gems and used version 1.0.0 of the debian base image


loburm:
Hi Andre,
We are in the process of deprecating this image. Please use gcr.io/stackdriver-agents/stackdriver-logging-agent instead. As far as I know it supposed to use latest fluentd version.

PR #253, opened 2019-04-10
An improvement of the distroless image path.

yuwenma:
Thanks @x13n for the original distroless change!


x13n:
LGTM, thanks for the contribution!

PR #254, opened 2019-04-11
Fix a Dockerfile linter

yuwenma:
Thanks @loburm for rebasing to distroless!


loburm:
Thank you for fixing it.

PR #259, opened 2019-05-07
Update scaler.sh

Crapworks:
Redirecting only stderr to /dev/null will spam the stackdriver logs with the acutal output of the "kubectl get" command. Both stdout and stderr should be redirected.

### DIFF
--- a/fluentd-gcp-scaler/scaler.sh
+++ b/fluentd-gcp-scaler/scaler.sh
@@ -23,7 +23,7 @@ log() {
 apply_scaling() {
   # This is assuming there is a ScalingPolicy installed in the cluster.
   # See https://github.com/justinsb/scaler for more details.
-  if ! kubectl get scalingpolicies -n ${NAMESPACE} ${SCALING_POLICY} &> /dev/null
+  if ! kubectl get scalingpolicies -n ${NAMESPACE} ${SCALING_POLICY} >/dev/null 2>&1
   then
     return
   fi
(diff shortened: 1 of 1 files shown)


kawych:
LGTM, adding Daniel to comment just in case this logging was intentional. Is the logging volume that high?

Crapworks:
It is not extremely high, but completely unnecessary (and we had customers complaining about it, since they basically have to pay for those logs). It also makes it harder to find actual errors in the logs.

This also only occurs if there actually IS a scaling policy for fluentd. If there is none the command would error, which is redirected by the current "2>/dev/null". So I guess this is not intended.

fluentd-gcp-scaler/scaler.sh
```diff
@@ -23,7 +23,7 @@ log() {
 apply_scaling() {
   # This is assuming there is a ScalingPolicy installed in the cluster.
   # See https://github.com/justinsb/scaler for more details.
-  if ! kubectl get scalingpolicies -n ${NAMESPACE} ${SCALING_POLICY} 2> /dev/null
+  if ! kubectl get scalingpolicies -n ${NAMESPACE} ${SCALING_POLICY} &> /dev/null
```
x13n: This is `/bin/sh`, not Bash. I don't think we can use `&>` here, using `>/dev/null 2>&1` is probably safer.

Crapworks: Good point, I've changed it.

x13n:
Thanks for contributing!

PR #262, opened 2019-05-13
Fix missing component name issue.

loburm:


prometheus-to-sd/translator/metric_descriptor_cache.go
```diff
@@ -41,6 +41,7 @@ func NewMetricDescriptorCache(service *v3.Service, config *config.CommonConfig)
 		broken:      make(map[string]bool),
 		service:     service,
 		config:      config,
+		component:   config.SourceConfig.Component,
```
x13n: It seems `component` and `config` fields are redundant in this struct, why not just use `config.SourceConfig.Component`? If the answer is convenience, I'd rather have a getter function rather than a field for the component.

loburm: Done.

PR #264, opened 2019-05-22
Add support for UNTYPED metrics

### LINKED ISSUE #261: Support UNTYPED metrics
Prometheus has an "untyped" metric type: https://github.com/prometheus/client_model/blob/fd36f4220a901265f90734c3183c5f0c91daa0b8/ruby/lib/prometheus/client/model/metrics.pb.rb#L11

According to the docs[1], prometheus doesn't really care too much about types but in StackDriver they do matter. Currently attempting to translate any UNTYPED metrics fails because the type is not supported.

Significantly, prometheus node-exporter uses UNTYPED for some very desirable metrics.

Is there any objection to treating UNTYPED metrics as a GAUGE? This kind makes the fewest assumptions about the data points. 

[1]: https://prometheus.io/docs/instrumenting/writing_exporters/#types

droyo:
Fixes #261 . Interprets UNTYPED metrics as a GAUGE of type DOUBLE.

### DIFF
--- a/prometheus-to-sd/translator/translator.go
+++ b/prometheus-to-sd/translator/translator.go
@@ -287,6 +287,8 @@ func setValue(mType dto.MetricType, valueType string, metric *dto.Metric, point 
 	} else if mType == dto.MetricType_HISTOGRAM {
 		point.Value.DistributionValue = convertToDistributionValue(metric.GetHistogram())
 		point.ForceSendFields = append(point.ForceSendFields, "DistributionValue")
+	} else if mType == dto.MetricType_UNTYPED {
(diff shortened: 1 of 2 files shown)


prometheus-to-sd/translator/translator.go
```diff
@@ -391,6 +392,9 @@ func extractValueType(mType dto.MetricType, originalDescriptor *v3.MetricDescrip
 	if mType == dto.MetricType_HISTOGRAM {
 		return "DISTRIBUTION"
 	}
+	if mType == dto.MetricType_UNTYPED {
```
loburm: You should update also setValue function above with value extraction from UNTYPED metric

droyo: done.

PR #269, opened 2019-07-17
Add support for https and authentication to prometheus-to-sd

44past4:


rwrz:
Please, is possible to provide basic auth too?

44past4:
@rwrz I have added support for basic auth.

prometheus-to-sd/translator/prometheus.go
```diff
@@ -64,6 +65,27 @@ func getPrometheusMetrics(config *config.SourceConfig) (*PrometheusResponse, err
...
+		},
+	}
+	request, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, err
+	}
+	if len(auth.Username) > 0 {
+		request.SetBasicAuth(auth.Username, auth.Password)
+	} else if len(auth.Token) > 0 {
+		request.Header.Set("Authorization", "Bearer "+auth.Token)
```
loburm: please use fmt.Sprintf() instead of string concatenation

prometheus-to-sd/main.go
```diff
@@ -45,8 +46,16 @@ var (
 		"The resolution at which prometheus-to-sd will scrape metric descriptors from Stackdriver.")
 	apioverride = flag.String("api-override", "",
 		"The stackdriver API endpoint to override the default one used (which is prod).")
-	source = flags.Uris{}
-	podId  = flag.String("pod-id", "machine",
+	source    = flags.Uris{}
+	authToken = flag.String("auth-token", "",
```
loburm: What if instead of making it as a global flags, we move this to per source configuration? In such case if different components have different ways to authorize we can use different tokens.

prometheus-to-sd/translator/prometheus.go
44past4: Fixed.

prometheus-to-sd/main.go
44past4: Fixed.

PR #275, opened 2019-07-30
Allow specifying metadata from flags, for use when not on GCE.

G-Harmon:
To use application default credentials, a secret needs to be mounted with the json creds,
and exposed via GOOGLE_APPLICATION_CREDENTIALS environment variable.

Here are the relevant parts of the yaml I used.
    spec:
      containers:
      - args:
        - /monitor
        # service exposing gke-connect-agent
        - --source=gke-connect-agent:http://10.103.158.164:8080
        - --manual-project=YOUR-PROJECT
        - --zone-override=us-central1-a
        - --manual-cluster=minikube-cluster
        - --manual-instance=MyGkeConnectAgent
        env:
        - name: GOOGLE_APPLICATION_CREDENTIALS
          value: /etc/creds/creds-gcp.json
        volumeMounts:
        - mountPath: /etc/creds
          name: creds-gcp
          readOnly: true
      volumes:
      - name: creds-gcp
        secret:
          defaultMode: 420
          secretName: creds-gcp

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/k8s-stackdriver/275)

### DIFF
--- a/prometheus-to-sd/config/gce_config.go
+++ b/prometheus-to-sd/config/gce_config.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	gce "cloud.google.com/go/compute/metadata"
+	"github.com/golang/glog"
 )
 
 // GceConfig aggregates all GCE related configuration parameters.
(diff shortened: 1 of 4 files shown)


prometheus-to-sd/config/gce_config.go
```diff
@@ -17,10 +17,23 @@ limitations under the License.
...
+)
+
+var (
+	// ManualProject is a flag that allows a different project to be set when the
+	// GCE Metadata server is not available.
+	ManualProject = flag.String("manual-project", "",
+		"Custom GCP project to send metrics to. Can be used when running outside GCE.")
+	manualCluster = flag.String("manual-cluster", "",
+		"Custom cluster name to use. Useful when running outside GCE.")
+	manualInstance = flag.String("manual-instance", "",
```
kawych: Can you name the flag "node_name"? "instance" is GCE-specific term, while "node" is k8s term.

The old gke_container monitored resource requires GCE instance_id. prom-to-sd ignores this convention and uses instance_name instead. New monitored resources k8s_container, k8s_pod, k8s_node (which should be recommended) require node name.

prometheus-to-sd/main.go
```diff
@@ -113,7 +113,18 @@ func main() {
 		glog.Error(http.ListenAndServe(fmt.Sprintf(":%d", *debugPort), nil))
 	}()
 
-	client := oauth2.NewClient(context.Background(), google.ComputeTokenSource(""))
+	var client *http.Client
+
+	if *config.ManualProject != "" {
```
kawych: Let's be consistent about this condition (earlier you use *manualCluster != "" || *manualInstance != "" || *ManualProject != ""). I think we should just check the *manualProject != "" (just as here)
- If not set, assume running on GCE, but allow to overwrite all other parameters individually (e.g. if someone uses their own weird tooling to set up kubernetes cluster and have their own node names and cluster name that is not picked up by Metadata Server)
- If set, require all other flags / default to empty strings.

prometheus-to-sd/config/gce_config.go
```diff
@@ -17,10 +17,23 @@ limitations under the License.
...
 	"strings"
 
 	gce "cloud.google.com/go/compute/metadata"
+	"github.com/golang/glog"
+)
+
+var (
+	// ManualProject is a flag that allows a different project to be set when the
+	// GCE Metadata server is not available.
+	ManualProject = flag.String("manual-project", "",
```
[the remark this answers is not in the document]
G-Harmon: The hack/verify-flags script wants flags to use "-" (dash) rather than "_" (underscore). I'm guessing these should use "-" as well?

prometheus-to-sd/config/gce_config.go
```diff
@@ -35,6 +48,20 @@ type GceConfig struct {
...
 func GetGceConfig(zone string, monitoredResourceTypes string) (*GceConfig, error) {
+	if *manualCluster != "" || *manualInstance != "" || *ManualProject != "" {
+		glog.Infof("Using customized metadata")
+		return &GceConfig{
+			Project: *ManualProject,
+			// zone should come from the command line --zone-override flag.
+			Zone:    zone,
+			Cluster: *manualCluster,
+			// GCE ClusterLocation was empty.
+			ClusterLocation:        "",
```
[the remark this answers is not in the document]
G-Harmon: No, I don't think we talked about empty Cluster Location. I will add a flag. When I tested this out, I took the default for --monitored-resource-types, which is gke_container.

prometheus-to-sd/main.go
G-Harmon: okay, I can do that.

G-Harmon: okay, done. I think this is what you meant.

prometheus-to-sd/config/gce_config.go
```diff
@@ -35,39 +50,62 @@ type GceConfig struct {
 
 // GetGceConfig builds GceConfig based on the provided prefix and metadata server available on GCE.
 func GetGceConfig(zone string, monitoredResourceTypes string) (*GceConfig, error) {
+	if *ProjectFlag != "" {
+		glog.Infof("Using metadata all from flags")
```
kawych: I think it's worth to log a warning if any of the values: cluster name, cluster location, node name is empty - usually it won't be intended.

prometheus-to-sd/config/gce_config.go
```diff
@@ -17,10 +17,25 @@ limitations under the License.
...
 import (
+	"flag"
 	"fmt"
 	"strings"
 
 	gce "cloud.google.com/go/compute/metadata"
+	"github.com/golang/glog"
+)
+
+var (
```
kawych: I just noticed that a convention for prom-to-sd is to include all flags in main.go: https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/prometheus-to-sd/main.go

Can you do the same for these? Here for example a zone parameter is passed, but it comes from zone-override flag.

kawych:
Thanks a lot for your change! I have two more small comments, other that that LGTM.

@loburm, since I'm OOO for the whole next week, would you mind helping with merging this PR once the comments are addressed?

prometheus-to-sd/config/gce_config.go
G-Harmon: okay, done. I now use --node-name to override gce.InstanceName(), which I think is what you were suggesting.

prometheus-to-sd/config/gce_config.go
G-Harmon: sure, done.

prometheus-to-sd/main.go
```diff
@@ -113,7 +121,18 @@ func main() {
 		glog.Error(http.ListenAndServe(fmt.Sprintf(":%d", *debugPort), nil))
 	}()
 
-	client := oauth2.NewClient(context.Background(), google.ComputeTokenSource(""))
+	var client *http.Client
+
+	if *projectOverride != "" {
+		client, err = google.DefaultClient(context.Background(), "https://www.googleapis.com/auth/cloud-platform")
```
loburm: I'm a little bit unfamiliar with how authentication works here, but how this is supposed to work? Shouldn't we pass some token for correct authorization?

G-Harmon: This uses Application Default Credentials, taken from the environment variable. To use it, the user needs to make sure GOOGLE_APPLICATION_CREDENTIALS points to the path of a file with credentials. (this might be useful: https://godoc.org/golang.org/x/oauth2/google#DefaultTokenSource)

prometheus-to-sd/config/gce_config.go
```diff
@@ -34,40 +35,74 @@ type GceConfig struct {
 }
 
 // GetGceConfig builds GceConfig based on the provided prefix and metadata server available on GCE.
-func GetGceConfig(zone string, monitoredResourceTypes string) (*GceConfig, error) {
+func GetGceConfig(project, cluster, clusterLocation, zone, node string, monitoredResourceTypes string) (*GceConfig, error) {
+	if project != "" {
+		glog.Infof("Using metadata all from flags")
+		if cluster == "" {
+			glog.Warning("cluster name was not set. This can be set with --cluster-name")
```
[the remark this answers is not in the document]
G-Harmon: okay, done.

prometheus-to-sd/config/gce_config.go
```diff
@@ -34,40 +35,74 @@ type GceConfig struct {
...
+		}
+		if clusterLocation == "" {
+			glog.Warning("cluster location was not set. This can be set with --cluster-location")
```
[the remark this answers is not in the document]
G-Harmon: done.

PR #276, opened 2019-07-30
added new param to allow configurable stack driver sink url

zhenglol:
## Why
We need a way to make event exporter talks to a non-production StackDriver for the purpose of integration testing.

## Work Done
* Adding another param to `sink-opt` so that we can configure the target StackDriver base url, instead of using the [hardcoded one](https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/event-exporter/vendor/google.golang.org/api/logging/v2/logging-gen.go#L46)
* `dep ensure -update` the `google.golang.org/api/logging` so that we can take advantage of the [`NewService` method](https://godoc.org/google.golang.org/api/logging/v2#NewService)
  * Have to manually `dep ensure -update` a few other pkgs due to incompatibility resulted by the api bump
    * `github.com/golang/protobuf`
    * `golang.org/x/net`
    * `golang.org/x/text`
    * `golang.org/x/oauth2`
  * everything under `vendor/` and `Gopkg.lock` are being modified automatically by `dep`

## Extra Thoughts
Current logging lib [is DEPRECATED](https://godoc.org/google.golang.org/api/logging/v2), however, the [new logging lib](https://godoc.org/cloud.google.com/go/logging) uses a different pattern while interacting with logger
* The concept of [`Service`](https://godoc.org/google.golang.org/api/logging/v2#Service) is being replaced by [`Client`](https://godoc.org/cloud.google.com/go/logging#NewClient) and [`Logger`](https://godoc.org/cloud.google.com/go/logging#Logger)
Being said, migrating to new client lib is out of the scope of my task here so the old client lib is simply being bumped and kept.


event-exporter/sinks/stackdriver/sink_factory.go
```diff
@@ -52,6 +53,7 @@ func NewSdSinkFactory() sinks.SinkFactory {
 			"concurrent requests to Stackdriver"),
 		resourceModelVersion: fs.String("stackdriver-resource-model", "", "Stackdriver resource model "+
 			"to be used for exports"),
+		endpoint: fs.String("endpoint", defaultEndpoint, "Stackdriver base url"),
```
loburm: Please update help string to: Base path for Stackdriver API

zhenglol: Updated, CI is green. Thanks! :+1:

loburm:
LGTM

PR #279, opened 2019-08-09
Add latency metric to prometheus-to-sd.

loburm:
It will help to estimate delay between moment when metric was scraped
and when it was send to the Stackdriver.

### DIFF
--- a/prometheus-to-sd/translator/stackdriver.go
+++ b/prometheus-to-sd/translator/stackdriver.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/golang/glog"
 	v3 "google.golang.org/api/monitoring/v3"
@@ -33,7 +34,7 @@ const (
 )
 
 // SendToStackdriver sends http request to Stackdriver to create the given timeseries.
-func SendToStackdriver(service *v3.Service, config *config.CommonConfig, ts []*v3.TimeSeries) {
+func SendToStackdriver(service *v3.Service, config *config.CommonConfig, ts []*v3.TimeSeries, scrapeTimestamp time.Time) {
 	if len(ts) == 0 {
 		glog.V(3).Infof("No metrics to send to Stackdriver for component %v", config.SourceConfig.Component)
 		return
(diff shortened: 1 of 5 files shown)


prometheus-to-sd/translator/stackdriver.go
```diff
@@ -53,15 +54,21 @@ func SendToStackdriver(service *v3.Service, config *config.CommonConfig, ts []*v
 			defer wg.Done()
 			req := &v3.CreateTimeSeriesRequest{TimeSeries: ts[begin:end]}
 			_, err := service.Projects.TimeSeries.Create(proj, req).Do()
+			now := time.Now()
```
serathius: What are you expecting to measure when error occurs?
E.g. SD is down, 500 will be returned imminently. This can hide potential problem with latency of successful requests.

prometheus-to-sd/translator/stackdriver.go
```diff
@@ -53,15 +54,21 @@ func SendToStackdriver(service *v3.Service, config *config.CommonConfig, ts []*v
...
 			}
+			for i := begin; i < end; i++ {
+				metricIngestionLatency.WithLabelValues(config.SourceConfig.Component).Observe(now.Sub(scrapeTimestamp).Seconds())
+			}
 		}(i, end)
 	}
 	wg.Wait()
 	sentTs := uint32(len(ts)) - failedTs
 	glog.V(4).Infof("Successfully sent %v timeseries to Stackdriver for component %v", sentTs, config.SourceConfig.Component)
+	for i := 0; i < len(ts); i++ {
```
serathius: Is this loop missing code?

prometheus-to-sd/translator/stackdriver.go
loburm: It makes sense to measure latency only for delivered timeseries, for failed that's obviously infinite. Haven't noticed that functions is not returning after error.

prometheus-to-sd/translator/stackdriver.go
loburm: Removed.

loburm:
Thanks, for the comments. Could you take one more look?

PR #284, opened 2019-08-30
Add latency metric to kubelet-to-gcm.

loburm:
This will allow to measure time it took to process scraped metrics,
before they have been send to the Stackdriver.


cezarygerard:
LGTM

PR #286, opened 2019-09-15
Fix memory leak when collecting prometheus metrics.

GeniyX:
For each prometheus metrics request, a separate `http.Client` object is created. After the request is completed, the connections created during its operation remain for the possibility of their reuse. Which leads to the accumulation of goroutine and memory leak.
Instead of creating `http.Client` for each request, we can reuse once created object, similar to how it was done in `net/http` package.

### DIFF
--- a/prometheus-to-sd/translator/prometheus.go
+++ b/prometheus-to-sd/translator/prometheus.go
@@ -66,7 +66,7 @@ func getPrometheusMetrics(config *config.SourceConfig) (*PrometheusResponse, err
 }
 
 // Allow insecure https connections
-var client = &http.Client{
+var prometheusHttpClient = &http.Client{
 	Transport: &http.Transport{
 		TLSClientConfig: &tls.Config{
 			InsecureSkipVerify: true,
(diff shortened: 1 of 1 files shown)


prometheus-to-sd/translator/prometheus.go
```diff
@@ -65,15 +65,16 @@ func getPrometheusMetrics(config *config.SourceConfig) (*PrometheusResponse, err
...
 
-func doPrometheusRequest(url string, auth config.AuthConfig) (resp *http.Response, err error) {
-	// Allow insecure https connections
-	client := &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
-			},
+// Allow insecure https connections
+var client = &http.Client{
```
Monnoroch: Since this is a global variable, I would suggest a more elaborate name, such as `prometheusHttpClient`.

loburm:
Thanks for your PR, this issue is already fixed in 0.8.1. Please add me as a reviewer next time.

PR #291, opened 2019-10-18
Add default value from GCE metadata server, when resource label flags

shoucongc:
Default project_id, cluster_name, location, and instance_id (or node_name) labels to values from GCE metadata server.

### DIFF
--- a/prometheus-to-sd/main.go
+++ b/prometheus-to-sd/main.go
@@ -66,8 +66,13 @@ var (
 	projectOverride = flag.String("project-id", "",
 		"GCP project to send metrics to. If not set, defaults to value from GCE Metadata Server.")
 	monitoredResourceTypePrefix = flag.String("monitored-resource-type-prefix", "", "MonitoredResource type prefix, to be appended by 'container', 'pod', and 'node'.")
-	monitoredResourceLabels     = flag.String("monitored-resource-labels", "", "Manually specified MonitoredResource labels. It is in URL parameter format, like 'A=B&C=D&E=F'. When this field is specified, 'monitored-resource-type-prefix' is also required. To note: 'namespace-name', 'pod-name', and 'container-name' should not be provided in this flag and they will always be overwritten by values in other command line flags.")
-	omitComponentName           = flag.Bool("omit-component-name", true,
+	monitoredResourceLabels     = flag.String("monitored-resource-labels", "", "Manually specified MonitoredResource labels. "+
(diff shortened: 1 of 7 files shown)


prometheus-to-sd/translator/translator.go
```diff
@@ -482,9 +482,31 @@ func getMonitoredResourceFromLabels(config *config.CommonConfig, labels []*dto.L
...
+	}
+	if _, found := resourceLabels["cluster_name"]; !found {
+		resourceLabels["cluster_name"] = config.GceConfig.Cluster
+	}
+	if _, found := resourceLabels["location"]; !found {
+		resourceLabels["location"] = config.GceConfig.Zone
+	}
+
+	// When MonitoredResource type is not "k8s_*", default "instance_id" label to GCE instance name.
+	if prefix != "k8s_" {
```
x13n: Shouldn't this be:
```
if prefix == "" {
```

prometheus-to-sd/main.go
```diff
@@ -66,7 +66,12 @@ var (
 	projectOverride = flag.String("project-id", "",
 		"GCP project to send metrics to. If not set, defaults to value from GCE Metadata Server.")
 	monitoredResourceTypePrefix = flag.String("monitored-resource-type-prefix", "", "MonitoredResource type prefix, to be appended by 'container', 'pod', and 'node'.")
-	monitoredResourceLabels     = flag.String("monitored-resource-labels", "", "Manually specified MonitoredResource labels. It is in URL parameter format, like 'A=B&C=D&E=F'. When this field is specified, 'monitored-resource-type-prefix' is also required. To note: 'namespace-name', 'pod-name', and 'container-name' should not be provided in this flag and they will always be overwritten by values in other command line flags.")
+	monitoredResourceLabels     = flag.String("monitored-resource-labels", "", "Manually specified MonitoredResource labels. " +
+	"It is in URL parameter format, like 'A=B&C=D&E=F'. " +
```
x13n: These strings should be indented more than the enclosing statement.

prometheus-to-sd/translator/translator.go
shoucongc: This if loop is the default for types other than "k8s_*".
And it should use numeric instance id.
For "k8s_node", it should use instance name.

prometheus-to-sd/main.go
shoucongc: Done.

shoucongc:
Added Instance ID fetch for non "k8s_*" types. PTAL. Thanks!

PR #294, opened 2019-10-22
Fix goroutine leak.

loburm:
Reuse the same transport to avoid recreating it all the time.

### DIFF
--- a/prometheus-to-sd/translator/prometheus.go
+++ b/prometheus-to-sd/translator/prometheus.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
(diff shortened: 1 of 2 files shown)


prometheus-to-sd/translator/prometheus.go
```diff
@@ -66,13 +77,8 @@ func getPrometheusMetrics(config *config.SourceConfig) (*PrometheusResponse, err
 }
 
 func doPrometheusRequest(url string, auth config.AuthConfig) (resp *http.Response, err error) {
-	// Allow insecure https connections
 	client := &http.Client{
```
x13n: This still creates a new HTTP client for each request. Can we have the client initialized just once?

loburm: Done.

prometheus-to-sd/translator/prometheus.go
```diff
@@ -36,6 +36,20 @@ type PrometheusResponse struct {
 	rawResponse string
 }
 
+var prometheusClient *http.Client
+
+func init() {
+	// Copy a DefaultTransport that is used by default http client.
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	// Allow insecure https connections.
+	transport.TLSClientConfig = &tls.Config{
```
serathius: Should we introduce some timeouts here?
https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779

loburm: There are some timeouts already present in DefaultTransport (that's why I'm using instead of an empty Transport). 
Also have added timeout to the client. I think 30 seconds should be safe to set.

PR #297, opened 2019-10-23
Fix monitored resource label bug in Go map, and bump version to 0.8.2

shoucongc:


shoucongc:
Added unit test. PTAL. Thanks!

PR #301, opened 2019-12-20
Allow specifying custom resource type and labels per source config

x13n:


prometheus-to-sd/config/source_config.go
```diff
@@ -112,7 +119,28 @@ func parseSourceConfig(uri flags.Uri, podId, namespaceId string) (*SourceConfig,
...
 	}
-	return newSourceConfig(component, protocol, host, port, path, *auth, whitelisted, metricsPrefix, podConfig, whitelistedLabelsMap)
+	return newSourceConfig(component, protocol, host, port, path, *auth, whitelisted, metricsPrefix, podConfig, whitelistedLabelsMap, customResource, customLabels)
+}
+
+func getMap(v url.Values, name string) map[string]string {
+	n := len(name)
+	m := make(map[string]string)
+	for k, vals := range v {
+		if len(k) <= n {
```
44past4: I believe it should be n+2. Otherwise passing "customLabels" as url query parameter will break the check below.

x13n: Which check are you referring to? `k[n] == '['`? When `len(k) == n` (or less), we'll hit `continue` statement. When `len(k) == n+1`, both `k[:n]` and `k[n]` will work.

prometheus-to-sd/translator/translator_test.go
```diff
@@ -473,6 +473,58 @@ func TestGetMonitoredResourceFromLabels(t *testing.T) {
...
+					Zone:       "us-east1-a",
+					Cluster:    "test-cluster",
+					Instance:   "default-instance",
+					InstanceId: "123",
+				},
+				SourceConfig: &config.SourceConfig{
+					CustomResourceType: "resource_foo",
+					CustomLabels: map[string]string{
+						"foo":        "bar",
+						"project_id": "",
```
loburm: I think it would be better to add all default labels to this test case, to ensure that they are not removed accidentally.

prometheus-to-sd/config/source_config.go
```diff
@@ -112,7 +119,28 @@ func parseSourceConfig(uri flags.Uri, podId, namespaceId string) (*SourceConfig,
...
+			continue
+		}
+		if k[:n] == name && k[n] == '[' && k[len(k)-1] == ']' {
```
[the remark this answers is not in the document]
x13n: I prefer to use index to avoid making an error in regex :-)

But seriously, this is what I added tests for. If you think test coverage is insufficient here, let me know, I'll add more tests.

prometheus-to-sd/translator/translator_test.go
x13n: Ack, done.

prometheus-to-sd/README.md
```diff
@@ -52,6 +52,29 @@ E.g, if `monitored-resource-type-prefix` = "k8s_", metrics will be written to "k
...
+  `monitored-resource-type-prefix`)
+- `customLabel[<label_name>]` specifies the labels that are provided for given
+  `customResourceType`. If `customResourceType` is not specified, this field is
+  ignored. All the labels for custom resource type has to be explicitly listed
+  in the source config. Omitting the value for the following labels will cause
+  prometheus-to-sd to use a value from GCE config:
+  - `instance_id`
+  - `cluster_name`
+  - `project_id`
+  - `location`
```
44past4: Maybe it would be worth adding 'node_name' as well which is used by k8s_node right now.

44past4:
Apart from the 2 minor comments LGTM.

prometheus-to-sd/README.md
```diff
@@ -52,6 +52,29 @@ E.g, if `monitored-resource-type-prefix` = "k8s_", metrics will be written to "k
...
 
+From 0.9.0, custom resource type and labels can be specified per source config.
+Within the `source` URL, two new flags are supported:
+- `customResourceType` specifies the resource type override. When this flag is
+  present, it will override other resource type settings (e.g.
+  `monitored-resource-type-prefix`)
+- `customLabel[<label_name>]` specifies the labels that are provided for given
+  `customResourceType`. If `customResourceType` is not specified, this field is
+  ignored. All the labels for custom resource type has to be explicitly listed
+  in the source config. Omitting the value for the following labels will cause
```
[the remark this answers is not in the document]
x13n: Done.

prometheus-to-sd/README.md
x13n: Good point, done.

PR #311, opened 2020-02-12
Don't list entities, fixes memory problems

brett-elliott:
Pull request to mitigate OOMs, bump version to v0.10.1

### DIFF
--- a/custom-metrics-stackdriver-adapter/Makefile
+++ b/custom-metrics-stackdriver-adapter/Makefile
@@ -3,7 +3,7 @@ GOOS?=linux
 OUT_DIR?=build
 PACKAGE=github.com/GoogleCloudPlatform/k8s-stackdriver/custom-metrics-stackdriver-adapter
 PREFIX?=staging-k8s.gcr.io
-TAG = v0.10.0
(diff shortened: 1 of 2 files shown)


kawych:
Do you plan to add caching? I would prefer to submit a single PR replacing calls to SD with a simple cache instead of splitting it.

brett-elliott:
PR doing this will be sent out

CatherineF-dev:
I tried this fix.  It reduces both memory & cpu.


- Before this PR: `kubectl get --raw "/apis/custom.metrics.k8s.io/v1beta2"` returns ALL pod metrics from GCP monitoring for this project. 

- After this PR: `kubectl get --raw "/apis/custom.metrics.k8s.io/v1beta2"` returns empty, which doesn't affect HPA in my cluster.

CatherineF-dev:
This is life of a HPA request and confirm that this LIST is not used in HPA. So we can accept this PR.

Screenshot 2024-01-10 at 9 22 31 AM

juli4n:
This was fixed in https://github.com/GoogleCloudPlatform/k8s-stackdriver/pull/623. Closing.

PR #323, opened 2020-04-23
Use go modules for vendor management

serathius:
* Config generated with `go init
github.com/GoogleCloudPlatform/k8s-stackdriver/custom-metrics-stackdriver-adapter`
* Dependencies frozen using replace
* Vendor regenerated using `go mod vendor`, regeneration caused:
  * Removal of not needed files
  * Removal of non .go files
  * Changes of files permission
* Tested if removal of `LICENSE` files changes result of generating
licenses.
/cc @kawych


kawych:
LGTM

PR #325, opened 2020-04-24
Use gomodules for vendor management

serathius:
Procedure per project:
* Run `go mod init <project name>`
* Add replace rules using script `go mod edit -json | jq -r '.Require[] | "replace \(.Path) => \(.Path) \(.Version)"'`
* Run `go mod tidy`
* Run `go mod vendor`
* Add `-mod=vendor` flag to all `go build` and `go test` commands
* Run tests and build `make test` and `make`
* Remove godep files
/cc @kawych


kawych:
LGTM

Issue #329, opened 2020-04-30
Event-exporter won't stop if received SIGINT or SIGTERM signal


lixiang233:
In newSystemStopChannel func, if received a SIGINT or SIGTERM signal, only one item will be put into the stop channel , but there's three places waiting for this channel (main process/sink/watcher), only one will receive the stop signal, so the process will not exit.
`signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)`
`sig := <-c`
`glog.Infof("Received signal %s, terminating", sig.String())`
`ch <- struct{}{}`

PR #330, opened 2020-04-30
Close event-exporter's stop channel after system terminating signal.

lixiang233:
Just close the stop channel to make sure every goroutine will receive stop signal. Otherwise event-exporter won't exit if received system terminating signal
#329


lixiang233:
@loburm can u review?

PR #333, opened 2020-05-14
Expose standard http server metrics

kawych:
The metrics are already registered by API machinery, they only need to be exposed.


juli4n:
This was added already in https://github.com/GoogleCloudPlatform/k8s-stackdriver/pull/375. Closing.

PR #334, opened 2020-05-15
prometheus-to-sd : stop exposing /debug/expvars endpoint on all interfaces by default

cezarygerard:
prometheus-to-sd : stop exposing metrics on all interfaces by default, listen just on localhost

### DIFF
--- a/hack/verify-flags/known-flags.txt
+++ b/hack/verify-flags/known-flags.txt
@@ -23,6 +23,7 @@ configuration-name
 current-release-pr
 custom-error-service
 datasource
+debug-address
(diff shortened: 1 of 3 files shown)


x13n:
What is the rationale for changing the default to localhost? This is a backwards-incompatible change, so you should bump major version of the image.

cezarygerard:
/debug/expvars exposes command line args so I have added a flag to register it on dedicated interface and port by default.
/metrics stays on same interface:port

x13n:
LGTM, but please bump the version.

PR #339, opened 2020-06-03
Archive event-adapter and fluentd-gcp-image

x13n:
Common way to archive a project is marking it as such on GitHub. Since we have multiple projects within this repo and hence cannot mark it as archived, I decided to clarify which ones are no longer supported by moving them to a dedicated directory.

@amujumdar FYI


kawych:
LGTM

PR #349, opened 2020-06-24
Use default token source when no token URL or project override is provided

### LINKED ISSUE #66: Authentication issue with prometheus-to-sd
Trying to use prometheus-to-sd in my project (as shown in the example kube config) but keep getting this error:

```
Error while sending request to Stackdriver googleapi: Error 403: Request had insufficient authentication scopes., forbidden
```

I also tried adding auth credentials for a service account that has the "monitoring admin" and "metrics writer" roles, but no dice. This is how the kube config for my prometheus-to-sd container looks like:

```
       - name: prometheus-to-sd
          image: gcr.io/google-containers/prometheus-to-sd:v0.2.1
          ports:
            - name: profiler
              containerPort: 6060
          command:
            - /monitor
            - --source=kube-state-metrics:http://localhost:5000
            - --pod-id=$(POD_NAME)
            - --namespace-id=$(POD_NAMESPACE)
          volumeMounts:
            - name: prometheus-to-sd-cred
              mountPath: /etc/cred
              readOnly: true
          env:
            - name: GOOGLE_APPLICATION_CREDENTIALS
              value: /etc/cred/prometheus-to-sd-cred.json
            - name: POD_NAME
              valueFrom:
                fieldRef:
                  fieldPath: metadata.name
            - name: POD_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
```

xichen2020:
cc @amujumdar @x13n @44past4 

This PR proposes using `DefaultTokenSource` instead of `ComputeTokenSource` as the token source to fall back on when no explicit token URL or project override is specified. The implication of using [DefaultTokenSource](https://github.com/golang/oauth2/blob/cdc340f7c179dbbfa4afd43b7614e8fcadde4269/google/default.go#L43) is that it'll attempt to load credentials from the location specified via `GOOGLE_APPLICATION_CREDENTIALS` and a few others before falling back to `ComputeTokenSource` (if the application is running on GCE).

Fixes #66 .

### DIFF
--- a/prometheus-to-sd/main.go
+++ b/prometheus-to-sd/main.go
@@ -155,7 +155,11 @@ func main() {
 		}
 		glog.Infof("Created a client with the default credentials")
 	} else {
-		client = oauth2.NewClient(context.Background(), google.ComputeTokenSource(""))
+		ts, err := google.DefaultTokenSource(context.Background(), "")
+		if err != nil {
+			glog.Fatalf("Error creating default token source: %v", err)
+		}
+		client = oauth2.NewClient(context.Background(), ts)
 	}
 
 	stackdriverService, err := v3.New(client)


x13n:
Thanks for the PR!

xichen2020:
Thanks for the review @x13n ! Do you know when the next version of `prom-to-sd` will be released to pick up this change?

x13n:
There is no release train to catch, we release new versions when needed. If you'd like to see a new version with your change, can you also bump the version tag in Makefile?

xichen2020:
Ack, thanks!

Opened #356 to bump the version.

PR #356, opened 2020-07-07
Bump prom-to-sd version to use DefaultTokenSource by default

xichen2020:
cc @x13n 

Bump `prom-to-sd` version to pick up the changes in #349 to use `DefaultTokenSource` when no token URL or project override is provided.

If the change is deemed backward incompatible enough to justify a major version bump, happy to do so.


x13n:
Thanks, I think this is ok as minor version bump.

PR #365, opened 2020-08-04
Remove `component` label from received_entry_count metric

x13n:
This label isn't very useful and can have very high cardinality.


kawych:
LGTM

PR #369, opened 2020-08-24
Remove memory total/limit metrics

x13n:
There is no total memory rerported via Summary API so they were
incorrect anyway.

### DIFF
--- a/kubelet-to-gcm/monitor/kubelet/translate.go
+++ b/kubelet-to-gcm/monitor/kubelet/translate.go
@@ -54,12 +54,6 @@ var (
 		Name:       "kubernetes.io/container/cpu/core_usage_time",
 	}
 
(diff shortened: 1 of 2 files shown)


44past4:
From what I see we have used AvailableBytes to calculate the total/limit metrics. 
AvailableBytes is defined as "memory limit - workingSetBytes" so theoretically by adding WorkingSetBytes to it we should be able to get "memory limit".
Therefore do we want to remove those metrics because the usage of AvailableBytes was incorrect or because this value was unavailable in most cases?

x13n:
The former. I think this should be fine for per-container use case, but I'm not sure how to interpret it for per node metrics. (I.e. is it total memory or allocatable memory). I need to verify this. Until then, this PR is on hold.

PR #382, opened 2020-10-02
Use 10000 page size on listing events.

mborsz:
In large clusters, we see up to 1M events and fetching them using default 500 page size is impossible due to limited token expiration time.

/assign @serathius


serathius:
LGTM
/cc @kawych

PR #385, opened 2020-10-13
Update prometheus-to-sd to detect wrong metric kind

JacobX33:
Prometheus-to-sd should stop trying to push metric with known wrong metric type.


loburm:
LGTM

PR #392, opened 2020-11-17
Update the Dockerfile for stackdriver-metrics-adapter to apply capabi…

vinayakankugoyal:
…lities to adapter binary so that it can bind to privileged ports without needing to be run as root.


serathius:
I think this is overkill, I would prefer to change the default port to 4443

vinayakankugoyal:
> I think this is overkill, I would prefer to change the default port to 4443

Thanks will do! Closing this PR.

PR #394, opened 2020-12-17
Bump dependencies to k8s 1.20

kawych:


serathius:
LGTM, should we consider removing `vendor/` directory?

Issue #399, opened 2021-02-16
prometheus-to-sd: GKE metadata instance/name not available.


jharshman:
When running prometheus-to-sd in a GKE cluster with Workload Identity enabled. All requests sent to the GCE metadata service are re-routed to the GKE metadata service. This is true unless you are running prometheus-to-sd in host networking mode which is in my opinion not necessary and undesirable.

Because the GKE metadata service does not support the GCE metdata service's `instance/name` endpoint, prometheus-to-sd will enter a crashloop and emit the following log:
```
Failed to get GCE config: error while getting instance (node) name: metadata: GCE metadata "instance/name" not defined
```

This is caused by the following code:
https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/7879c5733880d694d6871d5272725b75d45a5efc/prometheus-to-sd/config/gce_config.go#L87-L92

In order to circumvent this, the `instance/hostname` endpoint can be utilized in the event of an error fetching `instance/name` to serve as the node name.
Something along the lines of:
```
	// instance/name endpoint is not available on the GKE metadata server.
	// Try GCE instance/name endpoint. If error, try instance/hostname.
	// If instance/hostname, remove domain to replicate instance/name.
	if node == "" {
		node, err = gce.InstanceName()
		if err != nil {
			node, err = gce.Hostname()
			if err != nil {
				return nil, fmt.Errorf("error while getting instance (node) name: %v", err)
			}
			node = strings.Split(node, ".")[0]
		}
	}
```

PR #400, opened 2021-02-16
Utilize hostname when instance/node unavailable.

jharshman:
When running in GKE with workload identity enabled, prometheus-to-sd
will enter a crashloop unless you run it with host-networking enabled
which will bypass the GKE metadata server. This solves that issue by
utilizing instance/hostname when instance/name isn't available.

Fixes: #399

### DIFF
--- a/prometheus-to-sd/config/gce_config.go
+++ b/prometheus-to-sd/config/gce_config.go
@@ -84,10 +84,17 @@ func GetGceConfig(project, cluster, clusterLocation, zone, node string) (*GceCon
 		}
 	}
 
+	// instance/name endpoint is not available on the GKE metadata server.
+	// Try GCE instance/name endpoint. If error, try instance/hostname.
+	// If instance/hostname, remove domain to replicate instance/name.
 	if node == "" {
 		node, err = gce.InstanceName()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


prometheus-to-sd/config/gce_config.go
```diff
@@ -84,10 +84,17 @@ func GetGceConfig(project, cluster, clusterLocation, zone, node string) (*GceCon
...
+	// If instance/hostname, remove domain to replicate instance/name.
 	if node == "" {
 		node, err = gce.InstanceName()
 		if err != nil {
-			return nil, fmt.Errorf("error while getting instance (node) name: %v", err)
+			node, err = gce.Hostname()
+			if err != nil {
+				return nil, fmt.Errorf("error while getting instance (node) name: %v", err)
+			}
+			node = strings.Split(node, ".")[0]
```
x13n, line 97: Can you log a warning here? In case instance name isn't actually a part of the hostname and someone tries to debug this.

jharshman: @x13n done :)

x13n (approved):
LGTM, deferring to @solodov to merge.

PR #408, opened 2021-03-29
Stop talking to kube-controller-manager from kubelet-to-gcm when port is set to 0

x13n:
This isn't really used and occassionally causes problems. Leaving the functionality for now to make rolling back easier.


loburm:
LGTM

PR #409, opened 2021-03-30
Prevent invalid timestamps in page_fault_count metric

x13n:
Right after container is started, kubelet returns time == start_time for
that metric, which causes Monitoring API to return 400 error.


x13n:
FWIW, I discovered that when testing 1.3.4 release.

loburm:
LGTM

PR #410, opened 2021-03-30
Prevent invalid timestamps in cpu usage metrics

x13n:
Turns out CPU metrics are also affected. Verified there are no other errors.


loburm (approved):
LGTM

PR #422, opened 2021-06-04
Add arm64 support to event-exporter

pwschuurman:


event-exporter/Makefile
```diff
@@ -28,10 +28,23 @@ test:
 	${ENVVAR} go test -mod=vendor ./...
 
 container:
-	docker build --pull -t ${PREFIX}/${IMAGE_NAME}:${TAG} .
+	docker buildx build . --pull -t ${PREFIX}/${IMAGE_NAME}:${TAG} --load
+
+container-multiarch:
+	docker buildx build . --pull -t ${PREFIX}/${IMAGE_NAME}-amd64:${TAG} --platform=linux/amd64 --load
```
liu-cong, line 34: Suggestion: Have a variable called ALL_ARCH and build containers for ALL_ARCH in stead of hard coding. With that I think you can just reuse the `container` target.

pwschuurman: Updated the build targets to reference architecture in a more extensible way

solodov (approved):

rsgowman (approved):

PR #425, opened 2021-06-15
Add arm64 support to prometheus-to-sd

pwschuurman:


prometheus-to-sd/Makefile
```diff
@@ -15,20 +15,48 @@
 all: build
 
 ENVVAR = GOOS=linux GOARCH=amd64 CGO_ENABLED=0
-PREFIX = staging-k8s.gcr.io
-TAG = v0.11.1
+IMAGE_NAME=prometheus-to-sd
+ALL_ARCH=amd64 arm64
+PREFIX ?= staging-k8s.gcr.io
+TAG ?= v0.11.1
```
solodov, line 21: does it make sense to change the tag since technically it's a multi-arch image now?

pwschuurman: True, that does make sense. Updated the minor version for this and #422

solodov (approved):

PR #435, opened 2021-09-04
Add duration cmd line parameters

edhaight:
- Added alignment period cmd line parameter.
- Added rate interval cmd line parameter.
- Refactored alignment period and rate interval
  to be fields of the server options struct.


edhaight:
@googlebot I signed it!

juli4n:
Thanks for the pull request. Are you still interested in making this contribution?

Could you elaborate a bit more on the motivations behind this change? e.g. how users will decide when to pick different values? Updating the documentation in this CL with instructions on how (and when) to use these new knobs would help.

PR #438, opened 2021-09-30
Bring custom metrics stackdriver adapter up to date

shuaich:
This pull request brings the following changes:

- Update and simplify direct dependencies
- Change -mod flag value to use readonly instead of vendor
- Upgrade rbac.authorization.k8s.io API version to v1
- replace go version 1.13 with 1.16 for build

Testing:
Deploy updated image of custom metrics stackdriver adapter to k8s cluster v1.22 and run e2e test.

- [Feature:StackdriverCustomMetrics](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/instrumentation/monitoring/custom_metrics_stackdriver.go#L57o)
2 out 2 test cases passed
- [Feature:StackdriverExternalMetrics](https://github.com/kubernetes/kubernetes/blob/38746752af6d0ec955f4511a727da37347da3b92/test/e2e/instrumentation/monitoring/custom_metrics_stackdriver.go#L87)
1 out 1 test case passed
- [Feature:CustomMetricsAutoscaling](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/autoscaling/custom_metrics_stackdriver_autoscaling.go#L55)
6 out 7 test cases passed, the [failed one](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/autoscaling/custom_metrics_stackdriver_autoscaling.go#L133) is caused by a separate issue, which will be addressed in a new pull request


serathius:
lgtm

kawych:
lgtm

custom-metrics-stackdriver-adapter/Dockerfile
```diff
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.13-alpine as builder
+FROM golang:1.16-alpine as builder
+ENV GO111MODULE=on
 WORKDIR ${GOPATH}/src/github.com/GoogleCloudPlatform/k8s-stackdriver/custom-metrics-stackdriver-adapter
 COPY . ./
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -o /adapter
+RUN CGO_ENABLED=0 GOOS=linux go build -mod=mod -o /adapter
```
solodov, line 19: -mod=mod means go will automatically update go.mod, I don't that's what we want. -mod=readonly seems like a better choice.

custom-metrics-stackdriver-adapter/Dockerfile
```diff
@@ -12,10 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.13-alpine as builder
+FROM golang:1.16-alpine as builder
+ENV GO111MODULE=on
```
solodov, line 16: This is redundant, that's the default behaviour in go 1.16

custom-metrics-stackdriver-adapter/Dockerfile
shuaich:
```
In Go 1.15 and lower, the -mod=mod flag was enabled by default, so updates were performed automatically. Since Go 1.16, the go command acts as if -mod=readonly were set instead: if any changes to go.mod are needed, the go command reports an error and suggests a fix.
```
Good point. Will use **readonly**

custom-metrics-stackdriver-adapter/Dockerfile
shuaich: Good catch, will remove `ENV GO111MODULE=on`

custom-metrics-stackdriver-adapter/Makefile
```diff
@@ -20,7 +20,7 @@ push: docker
 	gcloud docker -- push ${PREFIX}/custom-metrics-stackdriver-adapter:$(TAG)
 
 test: $(PKG)
-	CGO_ENABLED=0 go test -mod vendor ./...
+	CGO_ENABLED=0 go test -mod mod ./...
```
solodov, line 23: another mod leftover

custom-metrics-stackdriver-adapter/Dockerfile
```diff
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.13-alpine as builder
+FROM golang:1.16-alpine as builder
 WORKDIR ${GOPATH}/src/github.com/GoogleCloudPlatform/k8s-stackdriver/custom-metrics-stackdriver-adapter
 COPY . ./
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -o /adapter
+RUN CGO_ENABLED=0 GOOS=linux go build -mod=readonly -o /adapter
```
kwiesmueller, line 18: I don't think we should switch away from vendor here.
The vendor directory enables us to have reproducible and faster builds as it's avoiding pulling dependencies on every build.
See https://golang.org/ref/mod#build-commands for more details.

In this case, all readonly does is prevent the build process from changing go.mod/go.sum, but if a dependency is temporary or permanently unavailable the build will still fail.

To update the vendor directory one should run
```sh
go mod tidy
go mod vendor
```

kwiesmueller (changes requested):

solodov (approved):

custom-metrics-stackdriver-adapter/Dockerfile
shuaich: > The vendor directory enables us to have reproducible and faster builds as it's avoiding pulling dependencies on every build.

We have deterministic build behaviour with the help of go.sum file. It's faster for the first time for the local build.

> if a dependency is temporary or permanently unavailable

Go module proxy caches these packages (public packages) permanently. For details, search for `I removed a bad release from my repository but it still appears in the mirror, what should I do?` in https://proxy.golang.org/

A dependency package might be temporarily unavailable, I don't see it as a risk, especially for automated build.

One potential benefit of using vendor is that we could see the diff between third party packages update. I believe we can get these information from commit log/release notes.

custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
```diff
@@ -36,7 +36,7 @@ subjects:
   name: custom-metrics-stackdriver-adapter
   namespace: custom-metrics
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.uthorization.k8s.io/v1
```
serathius, line 39: typo

shuaich: This has been fixed: https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml#L39

PR #442, opened 2021-10-20
Increment received_entry_count during list.

sophieliu15:
Event exporter writes a log entry to Cloud Logging during list by calling
writer.Write(). This entry is not currently counted in the metric
received_entry_count.

### DIFF
--- a/event-exporter/sinks/stackdriver/sink.go
+++ b/event-exporter/sinks/stackdriver/sink.go
@@ -114,6 +114,8 @@ func (s *sdSink) OnDelete(*corev1.Event) {
 // receiving the first list of events.
 func (s *sdSink) OnList(list *corev1.EventList) {
 	if s.beforeFirstList {
+		receivedEntryCount.Inc()
+		
 		entry := s.logEntryFactory.FromMessage("Event exporter started watching. " +
 			"Some events may have been lost up to this point.")
 		s.writer.Write([]*sd.LogEntry{entry}, s.logName, s.sdResourceFactory.defaultResource)
--- a/event-exporter/sinks/stackdriver/sink_test.go
+++ b/event-exporter/sinks/stackdriver/sink_test.go
@@ -28,14 +28,13 @@ import (
 )
 
 type fakeSdWriter struct {
-	writeFunc func([]*sd.LogEntry, string, *sd.MonitoredResource) int
+	writeFunc func([]*sd.LogEntry, string, *sd.MonitoredResource)
 }
 
-func (w *fakeSdWriter) Write(entries []*sd.LogEntry, logName string, resource *sd.MonitoredResource) int {
+func (w *fakeSdWriter) Write(entries []*sd.LogEntry, logName string, resource *sd.MonitoredResource) {
 	if w.writeFunc != nil {
-		return w.writeFunc(entries, logName, resource)
+		w.writeFunc(entries, logName, resource)
 	}
-	return 0
 }
 
 const (
@@ -165,12 +164,11 @@ func createSinkWith(params map[string]interface{}) (c *sdSinkConfig, s *sdSink, 
 	c = createConfig(params)
 	q = make(chan struct{}, 2*c.MaxConcurrency)
 	w := &fakeSdWriter{
-		writeFunc: func([]*sd.LogEntry, string, *sd.MonitoredResource) int {
+		writeFunc: func([]*sd.LogEntry, string, *sd.MonitoredResource) {
 			q <- struct{}{}
 			if v, ok := params[blockingParamName]; ok && v.(bool) {
 				<-done
 			}
-			return 0
 		},
 	}
 


event-exporter/sinks/stackdriver/sink_test.go
```diff
@@ -28,14 +28,13 @@ import (
 )
 
 type fakeSdWriter struct {
-	writeFunc func([]*sd.LogEntry, string, *sd.MonitoredResource) int
+	writeFunc func([]*sd.LogEntry, string, *sd.MonitoredResource)
```
rsgowman, line 31: Good catch. Although, I suspect this couldn't have even compiled previously. Is our CI system not checking this? It may be worth while to see if something is misconfigured.

event-exporter/sinks/stackdriver/sink.go
```diff
@@ -114,6 +114,8 @@ func (s *sdSink) OnDelete(*corev1.Event) {
 // receiving the first list of events.
 func (s *sdSink) OnList(list *corev1.EventList) {
 	if s.beforeFirstList {
+		receivedEntryCount.Inc()
```
rsgowman, line 117: Technically, we didn't receive this entry; we generated it. You could argue that the existing code is WAI. If we submit this change, this means that this will no longer match up with the number of events from k8s. Is it more worthwhile to match with that or with the events uploaded to stackdriver? (It's only one event either way, so unlikely to make much difference in practice.)

Any other thoughts?

(If we ultimately decide not to submit this PR, then we should at least add a comment here explaining this ^ to indicate that we're deliberately not incrementing the metric.)

sophieliu15: Looks like I didn't receive your comments before I submitted the PR.  :-)

From the definition of the metrics, it wants to include all the entries we plan to send to Stackdriver, so I think including the entry here is expected. From implementation point of view, it is easier counting this because it is more difficult to parse the entries that we send to Stackdriver successfully and then exclude this entry from the count for [successfullySentEntryCount](https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/event-exporter/sinks/stackdriver/writer.go#L86). 

For calculating log loss, I think it is ok to calculate log losses as (logs that are sent successfully) / (logs that we plan to send) regardless where the logs are coming from? I don't think it makes too much difference in practice as  you mentioned because it does not happen often (it happened 2 times in 30 days in one of my clusters).

Another way is to not generate a logs/events from this message, instead record it as a regular log. This is the only place in the code that actually generates a logs/events from a message instead of converting a Kubernetes event to a logs/events. Thinking about it, I actually prefer this option because in this way, everything Event Exporter exports as logs/events are actually coming from Kubernetes events.

WDYT?

rsgowman: > Looks like I didn't receive your comments before I submitted the PR. :-)
> 
> From the definition of the metrics, it wants to include all the entries we plan to send to Stackdriver, so I think including the entry here is expected. From implementation point of view, it is easier counting this because it is more difficult to parse the entries that we send to Stackdriver successfully and then exclude this entry from the count for [successfullySentEntryCount](https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/event-exporter/sinks/stackdriver/writer.go#L86).
> 
> For calculating log loss, I think it is ok to calculate log losses as (logs that are sent successfully) / (logs that we plan to send) regardless where the logs are coming from?

sgtm.

> I don't think it makes too much difference in practice as you mentioned because it does not happen often (it happened 2 times in 30 days in one of my clusters).

Yeah, if I'm reading the code correctly, I think this only happens once, on a restart. Assuming any non-trivial number k8s events on a cluster, that off-by-potentially-one issue in our measurements should be a rounding error.

> Another option is to not generate a logs/events from this message, instead record it as a regular log. This is the only place in the code that actually generates a logs/events from a message instead of converting a Kubernetes event to a logs/events. Thinking about it, I actually prefer this option because in this way, everything Event Exporter exports as logs/events are actually coming from Kubernetes events.

I like that approach too. This way, both sets of numbers line up nicely. One potential issue with that though, is that (I assume!) the log message would then be effectively out-of-band with respect to the events. In particular, the statement "Some events may have been lost up to this point" might end up in the "wrong" spot. Is this a concern?

rsgowman: > I was planning to use glog to record this message, similar to how other logs inside Event Exporter are handled. I assume this means at the end the log name in Cloud Logging for this message will change from logs/events to logs/stdout or logs/stderr? Is this what "wrong" spot you are referring to? Is the concern that some customers setting monitoring rules to this particular event so that after the change, the rule will no longer detect this event?

I was actually referring to a timing issue. eg imagine just writing this log to stdout and allowing fluentbit to pick it up. It'll eventually make its way to stackdriver regardless, but the order may be changed wrt the events. (Eg if fluentbit is a little slow, then this log would arrive *after* the event-exporter had already started to write events.)

Re logs/events -> logs/stdout, yeah, you're right, this would technically be a breaking change, though a pretty minor one. I suspect adding a quick release note entry would be sufficient.

> I think this change is closer to what Event Exporter intends to do. However, it will be a user visible change. I am not sure what is the guidelines here. If it involves a non-trivial release process (e.g., add a note to GKE release note etc), considering the product is going to retire soon and the impact is not much, we can just keep it as it is?

+1. I think keeping this as is, is the right decision.

PR #449, opened 2021-11-10
Support k8s.io/api/events/v1 API in Event Exporter.

sophieliu15:
LastTimestamp that is used in Event Exporter for log entry
timestamp field is deprecated in k8s.io/api/events/v1.

We will set timestamp of a log entry in the following order:
- Use LastTimestamp if presents. This means the event is emitted
using k8s.io/api/core/v1 library.
- Use Series.LastObservedTime if presents. This means the event
is emitted using k8s.io/api/events/v1 library.
- Use EventTime if presents. EventTime is the second best choice
when neither LastTimestamp nor Series.LastObservedTime presents.
- Use current time if none of the timestamps above presents.


event-exporter/sinks/stackdriver/log_entry_factory.go
```diff
@@ -36,8 +35,6 @@ var (
 	fieldBlacklist = []string{
 		// Is unnecessary, because it's demuxed already
 		"count",
-		// Timestamp is in the logEntry's metadata
-		"lastTimestamp",
```
sophieliu15, line 40: Review note: Now that we might use any of lastTimestamp, lastObservedTimestamp, eventTime, or current time as log entry timestamp, I don't see the need to exclude lastTimestamp from log payload. We can probably dynamically add the field that we use to set log entry timestamp to this list but the logic will be more complicated than what we are doing today. Also I don't see the need to exclude the field just because we use it to set log entry timestamp. Therefore I removed this field from the list in this commit. Feel free to let me know if you have any concerns/suggestions!

Note that this will require a note in GKE release note since it is a user visible change.

osalau: we'll also have to be transparent about this new behaviour of what the "timestamp" is in the release. Otherwise, exposing this field sgtm.

osalau (approved):

event-exporter/sinks/stackdriver/log_entry_factory.go
sophieliu15: Yes agreed. I think this PR needs a release notes whether we expose this field or not because it is a user visible change.

PR #461, opened 2022-03-04
Add a metric to measure latency between log creation and delivery.

sophieliu15:


event-exporter/sinks/stackdriver/metrics.go
```diff
@@ -36,3 +51,37 @@ func init() {
...
+		samples: make([]time.Time, 0),
+	}
+}
+
+func (lm *logTimestampBatch) append(s time.Time) {
+	lm.samples = append(lm.samples, s)
+}
+
+func measureLatencyOnSuccess(entries []*sd.LogEntry) {
+	timestampBatch := newLogTimestampBatch()
```
rsgowman, line 70: I'm not sure the `logTimestampBatch` type is providing much more value than just using a `[]time.Time` directly. If you got rid of it, I think this function would stay the same size and be just as readable. Consider doing so.

OTOH, you may want to keep it anyways if you feel it does improve readability, or if you have plans to extend it in the future.

rsgowman (approved):

event-exporter/sinks/stackdriver/metrics.go
sophieliu15: Thanks for the suggestion! Will remove the struct and use `[]time.Time` for now. If we plan to extend it in future, will consider making it a struct again.

rsgowman (approved):
lgtm.

PR #475, opened 2022-06-04
partial success on event-exporter writing logs

snowhork:
We should enable partial success on writing log entries.

In the for loop, some entries are in bad format, and it receives StatusBadRequest. But, if partials success is not enabled, some appropriate entries may lose.


snowhork:
@osalau @sophieliu15 please review.

event-exporter/sinks/stackdriver/writer.go
```diff
@@ -48,9 +48,10 @@ func newSdWriter(service *sd.Service) sdWriter {
 // unless the API returns BadRequest error.
 func (w sdWriterImpl) Write(entries []*sd.LogEntry, logName string, resource *sd.MonitoredResource) {
 	req := &sd.WriteLogEntriesRequest{
-		Entries:  entries,
-		LogName:  logName,
-		Resource: resource,
+		Entries:        entries,
+		LogName:        logName,
+		Resource:       resource,
+		PartialSuccess: true,
```
sophieliu15, line 54: Thanks for sending in the PR!

We also need to count how many entries are successfully delivered in our metrics here and their corresponding latency: https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/event-exporter/sinks/stackdriver/writer.go#L66-L67

If there is an easy way to count those, could you update corresponding metrics? If not, feel free to send us a feature request PR and we will prioritize the work based on our team capacity and priority.

snowhork: @sophieliu15 Thanks for your reply! I will try it.

PR #476, opened 2022-06-06
Increase memory for adapter

chajath:
Adapter has been constantly OOM killed. With the increased memory, things seem to be stable.


CatherineF-dev:
/lgtm

cc @erain to approve

juli4n:
Thanks for the PR.

I'm a bit reluctant to do a blanket update of the deployment request/limits unless we prove the adapter strictly requires that to be functional. Note that there have been recent changes[1] to make the adapter consume less memory.

As the adapter is installed and managed by end-users, they can always adapt the request / limits as needed.

[1]: https://github.com/GoogleCloudPlatform/k8s-stackdriver/pull/623

PR #487, opened 2022-10-17
[Feature] Support Prometheus Metrics Stored in Google Cloud Monitoring

bryan-hz:
Example use case for HPA from the [example](/GoogleCloudPlatform/k8s-stackdriver/tree/master/custom-metrics-stackdriver-adapter/examples/prometheus-to-sd):
Replacing the yaml file with
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    run: custom-metric-prometheus-sd
  name: custom-metric-prometheus-sd
  namespace: {{ NAMESAPCE }}
spec:
  replicas: 1
  selector:
    matchLabels:
      run: custom-metric-prometheus-sd
  template:
    metadata:
      labels:
        run: custom-metric-prometheus-sd
    spec:
      containers:
      - command: ["./prometheus-dummy-exporter"]
        args:
        - --metric-name=foo
        - --metric-value=40
        - --port=8080
        image: gcr.io/google-samples/prometheus-dummy-exporter:v0.2.0
        name: prometheus-dummy-exporter
        resources:
          requests:
            cpu: 100m
---
apiVersion: monitoring.googleapis.com/v1
kind: PodMonitoring
metadata:
  name: prom-example
  namespace: {{ NAMESPACE }}
spec:
  selector:
    matchLabels:
      run: custom-metric-prometheus-sd
  endpoints:
  - port: 8080
    interval: 30s
---
apiVersion: autoscaling/v2beta2
kind: HorizontalPodAutoscaler
metadata:
  name: custom-metric-prometheus-sd
  namespace: {{ NAMESPACE }}
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: custom-metric-prometheus-sd
  minReplicas: 1
  maxReplicas: 5
  metrics:
  - type: Pods
    pods:
      metric:
        name: prometheus.googleapis.com|foo|gauge
      target:
        type: AverageValue
        averageValue: 20
```

### DIFF
--- a/custom-metrics-stackdriver-adapter/adapter.go
+++ b/custom-metrics-stackdriver-adapter/adapter.go
@@ -148,10 +148,10 @@ func main() {
 	flags.AddGoFlagSet(flag.CommandLine) // make sure we get the klog flags
 
 	serverOptions := stackdriverAdapterServerOptions{
-		UseNewResourceModel:         false,
+		UseNewResourceModel:         true,
 		EnableCustomMetricsAPI:      true,
 		EnableExternalMetricsAPI:    true,
-		FallbackForContainerMetrics: false,
+		FallbackForContainerMetrics: true,
 		EnableCoreMetricsAPI:        false,
 		EnableDistributionSupport:   false,
 	}
--- a/custom-metrics-stackdriver-adapter/adapter.go
+++ b/custom-metrics-stackdriver-adapter/adapter.go
@@ -148,10 +148,10 @@ func main() {
 	flags.AddGoFlagSet(flag.CommandLine) // make sure we get the klog flags
 
 	serverOptions := stackdriverAdapterServerOptions{
-		UseNewResourceModel:         true,
+		UseNewResourceModel:         false,
 		EnableCustomMetricsAPI:      true,
 		EnableExternalMetricsAPI:    true,
-		FallbackForContainerMetrics: true,
+		FallbackForContainerMetrics: false,
 		EnableCoreMetricsAPI:        false,
 		EnableDistributionSupport:   false,
 	}
--- a/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
+++ b/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
@@ -45,9 +45,13 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - "*"
+  - pods
+  - nodes
+  - nodes/stats
   verbs:
-  - "*"
+  - get
+  - list
+  - watch
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
+++ b/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
@@ -37,13 +37,26 @@ subjects:
   namespace: custom-metrics
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: custom-metrics-resource-reader
+  namespace: custom-metrics
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - "*"
+  verbs:
+  - "*"
+---
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: custom-metrics-resource-reader
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: view
+  name: custom-metrics-resource-reader
 subjects:
 - kind: ServiceAccount
   name: custom-metrics-stackdriver-adapter
--- a/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
+++ b/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
@@ -77,8 +77,6 @@ spec:
         name: pod-custom-metrics-stackdriver-adapter
         command:
         - /adapter
-        - --use-new-resource-model=true
-        - --fallback-for-container-metrics=true
         resources:
           limits:
             cpu: 250m
--- a/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
+++ b/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
@@ -77,6 +77,8 @@ spec:
         name: pod-custom-metrics-stackdriver-adapter
         command:
         - /adapter
+        - --use-new-resource-model=true
+        - --fallback-for-container-metrics=true
         resources:
           limits:
             cpu: 250m
--- a/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
+++ b/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
@@ -165,14 +165,23 @@ func (p *StackdriverProvider) getNamespacedMetricByName(groupResource schema.Gro
 		return nil, NewOperationNotSupportedError(fmt.Sprintf("Get namespaced metric by name for resource %q", groupResource.Resource))
 	}
 	matchingPod, err := p.kubeClient.Pods(namespace).Get(context.Background(), name, metav1.GetOptions{})
+	pods := &v1.PodList{Items: []v1.Pod{*matchingPod}}
 	if err != nil {
 		return nil, err
 	}
-	metricKind, metricValueType, err := p.translator.GetMetricKind(getCustomMetricName(escapedMetricName), metricSelector)
+	metricName := getCustomMetricName(escapedMetricName)
+	metricKind, metricValueType, err := p.translator.GetMetricKind(metricName, metricSelector)
 	if err != nil {
 		return nil, err
 	}
-	stackdriverRequest, err := p.translator.GetSDReqForPods(&v1.PodList{Items: []v1.Pod{*matchingPod}}, getCustomMetricName(escapedMetricName), metricKind, metricValueType, metricSelector, namespace)
+	queryBuilder := translator.NewQueryBuilder(p.translator, metricName)
+	stackdriverRequest, err := queryBuilder.
+		WithPods(pods).
+		WithMetricKind(metricKind).
+		WithMetricValueType(metricValueType).
+		WithMetricSelector(metricSelector).
+		WithNamespace(namespace).
+		Build()
 	if err != nil {
 		return nil, err
 	}
@@ -182,7 +191,7 @@ func (p *StackdriverProvider) getNamespacedMetricByName(groupResource schema.Gro
 	}
 
 	if p.fallbackForContainerMetrics && len(stackdriverResponse.TimeSeries) == 0 {
-		stackdriverRequest, err = p.translator.GetSDReqForContainers(&v1.PodList{Items: []v1.Pod{*matchingPod}}, getCustomMetricName(escapedMetricName), metricKind, metricValueType, metricSelector, namespace)
+		stackdriverRequest, err = p.translator.GetSDReqForContainers(pods, metricName, metricKind, metricValueType, metricSelector, namespace)
 		if err != nil {
 			return nil, err
 		}
@@ -209,15 +218,25 @@ func (p *StackdriverProvider) getNamespacedMetricBySelector(groupResource schema
 	if err != nil {
 		return nil, err
 	}
-	metricKind, metricValueType, err := p.translator.GetMetricKind(getCustomMetricName(escapedMetricName), metricSelector)
+	metricName := getCustomMetricName(escapedMetricName)
+	klog.V(4).Infof("Querying for metric: %s", metricName)
+
+	metricKind, metricValueType, err := p.translator.GetMetricKind(metricName, metricSelector)
 	if err != nil {
 		return nil, err
 	}
+
+	queryBuilder := translator.NewQueryBuilder(p.translator, metricName).
+		WithMetricKind(metricKind).
+		WithMetricSelector(metricSelector).
+		WithMetricValueType(metricValueType).
+		WithNamespace(namespace)
+
 	result := []custom_metrics.MetricValue{}
 	for i := 0; i < len(matchingPods.Items); i += translator.MaxNumOfArgsInOneOfFilter {
 		sliceSegmentEnd := min(i+translator.MaxNumOfArgsInOneOfFilter, len(matchingPods.Items))
 		podsSlice := &v1.PodList{Items: matchingPods.Items[i:sliceSegmentEnd]}
-		stackdriverRequest, err := p.translator.GetSDReqForPods(podsSlice, getCustomMetricName(escapedMetricName), metricKind, metricValueType, metricSelector, namespace)
+		stackdriverRequest, err := queryBuilder.WithPods(podsSlice).Build()
 		if err != nil {
 			return nil, err
 		}
--- a/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
+++ b/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
@@ -195,7 +195,14 @@ func (p *StackdriverProvider) getNamespacedMetricByName(groupResource schema.Gro
 	}
 
 	if p.fallbackForContainerMetrics && len(stackdriverResponse.TimeSeries) == 0 {
-		stackdriverRequest, err = p.translator.GetSDReqForContainers(pods, metricName, metricKind, metricValueType, metricSelector, namespace)
+		stackdriverRequest, err := queryBuilder.
+			AsContainerType().
+			WithPods(pods).
+			WithMetricKind(metricKind).
+			WithMetricValueType(metricValueType).
+			WithMetricSelector(metricSelector).
+			WithNamespace(namespace).
+			Build()
 		if err != nil {
 			return nil, err
 		}
@@ -259,7 +266,14 @@ func (p *StackdriverProvider) getNamespacedMetricBySelector(groupResource schema
 		for i := 0; i < len(matchingPods.Items); i += translator.MaxNumOfArgsInOneOfFilter {
 			sliceSegmentEnd := min(i+translator.MaxNumOfArgsInOneOfFilter, len(matchingPods.Items))
 			podsSlice := &v1.PodList{Items: matchingPods.Items[i:sliceSegmentEnd]}
-			stackdriverRequest, err := p.translator.GetSDReqForContainers(podsSlice, getCustomMetricName(escapedMetricName), metricKind, metricValueType, metricSelector, namespace)
+			stackdriverRequest, err := translator.NewQueryBuilder(p.translator, metricName).
+				AsContainerType().
+				WithPods(podsSlice).
+				WithMetricKind(metricKind).
+				WithMetricValueType(metricValueType).
+				WithMetricSelector(metricSelector).
+				WithNamespace(namespace).
+				Build()
 			if err != nil {
 				return nil, err
 			}
--- a/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
+++ b/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
@@ -105,11 +105,17 @@ func (p *StackdriverProvider) getRootScopedMetricByName(groupResource schema.Gro
 	if err != nil {
 		return nil, err
 	}
-	metricKind, metricValueType, err := p.translator.GetMetricKind(getCustomMetricName(escapedMetricName), metricSelector)
+	metricName := getCustomMetricName(escapedMetricName)
+	metricKind, metricValueType, err := p.translator.GetMetricKind(metricName, metricSelector)
 	if err != nil {
 		return nil, err
 	}
-	stackdriverRequest, err := p.translator.GetSDReqForNodes(&v1.NodeList{Items: []v1.Node{*matchingNode}}, getCustomMetricName(escapedMetricName), metricKind, metricValueType, metricSelector)
+	stackdriverRequest, err := translator.NewQueryBuilder(p.translator, metricName).
+		WithNodes(&v1.NodeList{Items: []v1.Node{*matchingNode}}).
+		WithMetricKind(metricKind).
+		WithMetricValueType(metricValueType).
+		WithMetricSelector(metricSelector).
+		Build()
 	if err != nil {
 		return nil, err
 	}
@@ -133,7 +139,8 @@ func (p *StackdriverProvider) getRootScopedMetricBySelector(groupResource schema
 	if err != nil {
 		return nil, err
 	}
-	metricKind, metricValueType, err := p.translator.GetMetricKind(getCustomMetricName(escapedMetricName), metricSelector)
+	metricName := getCustomMetricName(escapedMetricName)
+	metricKind, metricValueType, err := p.translator.GetMetricKind(metricName, metricSelector)
 	if err != nil {
 		return nil, err
 	}
@@ -141,7 +148,12 @@ func (p *StackdriverProvider) getRootScopedMetricBySelector(groupResource schema
 	for i := 0; i < len(matchingNodes.Items); i += translator.MaxNumOfArgsInOneOfFilter {
 		sliceSegmentEnd := min(i+translator.MaxNumOfArgsInOneOfFilter, len(matchingNodes.Items))
 		nodesSlice := &v1.NodeList{Items: matchingNodes.Items[i:sliceSegmentEnd]}
-		stackdriverRequest, err := p.translator.GetSDReqForNodes(nodesSlice, getCustomMetricName(escapedMetricName), metricKind, metricValueType, metricSelector)
+		stackdriverRequest, err := translator.NewQueryBuilder(p.translator, metricName).
+			WithNodes(nodesSlice).
+			WithMetricKind(metricKind).
+			WithMetricValueType(metricValueType).
+			WithMetricSelector(metricSelector).
+			Build()
 		if err != nil {
 			return nil, err
 		}
--- a/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
+++ b/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
@@ -164,17 +164,20 @@ func (p *StackdriverProvider) getNamespacedMetricByName(groupResource schema.Gro
 	if groupResource.Resource != podResource {
 		return nil, NewOperationNotSupportedError(fmt.Sprintf("Get namespaced metric by name for resource %q", groupResource.Resource))
 	}
+
 	matchingPod, err := p.kubeClient.Pods(namespace).Get(context.Background(), name, metav1.GetOptions{})
-	pods := &v1.PodList{Items: []v1.Pod{*matchingPod}}
 	if err != nil {
 		return nil, err
 	}
+
 	metricName := getCustomMetricName(escapedMetricName)
 	metricKind, metricValueType, err := p.translator.GetMetricKind(metricName, metricSelector)
 	if err != nil {
 		return nil, err
 	}
 	queryBuilder := translator.NewQueryBuilder(p.translator, metricName)
+
+	pods := &v1.PodList{Items: []v1.Pod{*matchingPod}}
 	stackdriverRequest, err := queryBuilder.
 		WithPods(pods).
 		WithMetricKind(metricKind).
@@ -182,6 +185,7 @@ func (p *StackdriverProvider) getNamespacedMetricByName(groupResource schema.Gro
 		WithMetricSelector(metricSelector).
 		WithNamespace(namespace).
 		Build()
+
 	if err != nil {
 		return nil, err
 	}
--- a/custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
+++ b/custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
@@ -98,6 +98,9 @@ type podsContainer struct {
 }
 
 // isPodsContainerValid checks if podsContainer is valid to be used
+//
+// this only happens when the both pods & podNames are provided,
+// for example, when WithPods() and WithPodNames() are both used in QueryBuilder
 func (pc podsContainer) isPodsContainerValid() bool {
 	return !(len(pc.podNames) > 0 && pc.pods != nil && len(pc.pods.Items) > 0)
 }
@@ -109,26 +112,26 @@ func (pc podsContainer) isPodsContainerEmpty() bool {
 
 // getPodNames gets pod names from podsContainer is any provided
 func (pc podsContainer) getPodNames() []string {
-	if pc.pods != nil {
-		podNames := []string{}
-		for _, item := range pc.pods.Items {
-			podNames = append(podNames, fmt.Sprintf("%q", item.GetName()))
-		}
-		return podNames
+	if pc.pods == nil {
+		return pc.podNames
 	}
-	return pc.podNames
+	podNames := make([]string, len(pc.pods.Items))
+	for i, item := range pc.pods.Items {
+		podNames[i] = fmt.Sprintf("%q", item.GetName())
+	}
+	return podNames
 }
 
-// getPodIds gets pod ids from podsContainer if any provided
-func (pc podsContainer) getPodIds() []string {
-	if pc.pods != nil {
-		podIds := []string{}
-		for _, item := range pc.pods.Items {
-			podIds = append(podIds, fmt.Sprintf("%q", item.GetUID()))
-		}
-		return podIds
+// getPodIDs gets pod ids from podsContainer if any provided
+func (pc podsContainer) getPodIDs() []string {
+	if pc.pods == nil {
+		return []string{}
 	}
-	return []string{}
+	podIDs := make([]string, len(pc.pods.Items))
+	for i, item := range pc.pods.Items {
+		podIDs[i] = fmt.Sprintf("%q", item.GetUID())
+	}
+	return podIDs
 }
 
 // nodesContainer is a helper struct to hold nodes values
@@ -138,6 +141,9 @@ type nodesContainer struct {
 }
 
 // isNodesContainerValid checks if nodeContainer is valid to be used
+//
+// this only happens when the both nodes & nodeNames are provided,
+// for example, when WithNodes() and WithNodeNames() are both used in QueryBuilder
 func (nc nodesContainer) isNodesContainerValid() bool {
 	return !(len(nc.nodeNames) > 0 && nc.nodes != nil && len(nc.nodes.Items) > 0)
 }
@@ -305,7 +311,7 @@ func (qb QueryBuilder) getResourceNames() []string {
 		return qb.nodes.getNodeNames()
 	}
 	// legacy resource model
-	return qb.pods.getPodIds()
+	return qb.pods.getPodIDs()
 }
 
 // AsContainerType enforces to query k8s_container type metrics
--- a/custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
+++ b/custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
@@ -135,12 +135,15 @@ func (qb *QueryBuilder) WithPods(pods *v1.PodList) *QueryBuilder {
 	return qb
 }
 
+// podList is required to be no longer than MaxNumOfArgsInOneOfFilter items. This is enforced by limitation of
+// "one_of()" operator in Stackdriver filters, see documentation:
+// https://cloud.google.com/monitoring/api/v3/filters
 func (qb *QueryBuilder) validate() error {
 	if len(qb.pods.Items) == 0 {
 		return apierr.NewBadRequest("No objects matched provided selector")
 	}
 	if len(qb.pods.Items) > MaxNumOfArgsInOneOfFilter {
-		return apierr.NewInternalError(fmt.Errorf("GetSDReqForPods called with %v pod list, but allowed limit is %v pods", len(qb.pods.Items), MaxNumOfArgsInOneOfFilter))
+		return apierr.NewInternalError(fmt.Errorf("QueryBuilder tries to build with %v pod list, but allowed limit is %v pods", len(qb.pods.Items), MaxNumOfArgsInOneOfFilter))
 	}
 	if qb.metricValueType == "DISTRIBUTION" && !qb.translator.supportDistributions {
 		return apierr.NewBadRequest("Distributions are not supported")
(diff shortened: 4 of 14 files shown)


bryan-hz:
@couragej

custom-metrics-stackdriver-adapter/adapter.go
```diff
@@ -148,10 +148,10 @@ func main() {
 	flags.AddGoFlagSet(flag.CommandLine) // make sure we get the klog flags
 
 	serverOptions := stackdriverAdapterServerOptions{
```
courageJ, line 150: Why do we change the default values?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -348,6 +414,7 @@ func quoteAll(list []string) []string {
 	return result
 }
 
+// Deprecated since FilterBuilder, use FilterBuilder instead
```
courageJ, line 417: Can we deprecate this function and replace all dependencies to use FilterBuilder?

bryan-hz: Yeah, eventually we will need to remove this function, but to give it a smooth deprecation process and avoid the risk of introduce breaking changes, I recommend mark it as deprecated and each time we only update those use cases we touch. 

Lastly, when all use cases of this function are replaced with `FilterBuilder`, then we remove this function (same for other deprecated function as well)

how do you think? 🤔

custom-metrics-stackdriver-adapter/adapter.go
courageJ: I was proposing to support prometheus_target by default without adding new flags.

bryan-hz: as discussed, I will revert this change

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -358,36 +425,43 @@ func joinFilters(filters ...string) string {
...
 func (t *Translator) filterForAnyNode() string {
 	return "resource.type = \"k8s_node\""
 }
 
+// Deprecated since FilterBuilder, use FilterBuilder instead
 func (t *Translator) filterForAnyContainer() string {
 	return "resource.type = \"k8s_container\""
 }
 
+// Deprecated since FilterBuilder, use FilterBuilder instead
```
kwiesmueller, line 456:
```suggestion
// Deprecated, use FilterBuilder instead
```
I think that's just as clear

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -0,0 +1,144 @@
...
+		"legacy":     LegacyType,
+	}
+)
+
+type FilterBuilder struct {
+	schema  *Schema
+	filters []string
+}
+
+// initialize with resource type and whether to use legacy model
```
kwiesmueller, line 67: Same nit regarding exported comments.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -0,0 +1,144 @@
...
+// For legacy pod, use NewFilterBuilder(AnyType, false)
+func NewFilterBuilder(resourceType string) *FilterBuilder {
+	var schema *Schema
+	switch resourceType {
+	case PrometheusType:
+		schema = PrometheusSchema
+	case LegacyType:
+		schema = LegacyPodSchema
+	default:
```
kwiesmueller, line 76: Can you make the PodSchema a dedicated too so it's clear that it's expected?
The default can stay as is, I'd just prefer all defined cases to be specified as such.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -0,0 +1,144 @@
...
+func (fb *FilterBuilder) WithNamespace(namespace string) *FilterBuilder {
+	if namespace != "" {
+		fb.filters = append(fb.filters, fmt.Sprintf("%s = %q", fb.schema.namespace, namespace))
+	}
+	return fb
+}
+
+func (fb *FilterBuilder) WithPods(pods []string) *FilterBuilder {
+	if len(pods) == 0 {
+		klog.Fatalf("createFilterForPods called with empty list of pod names")
```
kwiesmueller, line 128: Is this a reason to panic?
Does this indicate a bad response or a coding error? If it's just bad input data we need a better way of handling this as we can't just panic every time we get bad data.

kwiesmueller (changes requested):

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -88,58 +89,123 @@ type Translator struct {
...
-		return nil, apierr.NewInternalError(fmt.Errorf("GetSDReqForPods called with %v pod list, but allowed limit is %v pods", len(podList.Items), MaxNumOfArgsInOneOfFilter))
+	if len(qb.pods.Items) > MaxNumOfArgsInOneOfFilter {
+		return apierr.NewInternalError(fmt.Errorf("QueryBuilder tries to build with %v pod list, but allowed limit is %v pods", len(qb.pods.Items), MaxNumOfArgsInOneOfFilter))
 	}
-	if metricValueType == "DISTRIBUTION" && !t.supportDistributions {
-		return nil, apierr.NewBadRequest("Distributions are not supported")
+	if qb.metricValueType == "DISTRIBUTION" && !qb.translator.supportDistributions {
+		return apierr.NewBadRequest("Distributions are not supported")
 	}
+
```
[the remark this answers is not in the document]
bryan-hz: sounds good, will add it

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -88,58 +89,123 @@ type Translator struct {
...
-	if t.useNewResourceModel {
-		resourceNames := getPodNames(podList)
-		filter = joinFilters(
-			t.filterForMetric(metricName),
-			t.filterForCluster(),
-			t.filterForPods(resourceNames, namespace),
-			t.filterForAnyPod())
+	if qb.translator.useNewResourceModel {
+		var filterBuilder *utils.FilterBuilder
+		if strings.HasPrefix(qb.metricName, "prometheus.googleapis.com") {
```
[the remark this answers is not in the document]
bryan-hz: yeah...it's a little bit tricky here, probably we could come back and fix it in the future

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
bryan-hz: sounds good to me

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
bryan-hz: Gotcha, this comes from the original code before I do the refactoring, let me think up a better a way handling this

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -182,7 +191,7 @@ func (p *StackdriverProvider) getNamespacedMetricByName(groupResource schema.Gro
 	}
 
 	if p.fallbackForContainerMetrics && len(stackdriverResponse.TimeSeries) == 0 {
-		stackdriverRequest, err = p.translator.GetSDReqForContainers(&v1.PodList{Items: []v1.Pod{*matchingPod}}, getCustomMetricName(escapedMetricName), metricKind, metricValueType, metricSelector, namespace)
+		stackdriverRequest, err = p.translator.GetSDReqForContainers(pods, metricName, metricKind, metricValueType, metricSelector, namespace)
```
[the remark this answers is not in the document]
bryan-hz: done

bryan-hz:
Hi @courageJ @kwiesmueller , I resolved the comments in the latest 3 commits, please help review the changes when you get a chance, thanks in advance! 😸

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -89,100 +91,229 @@ type Translator struct {
...
 
-	var filter string
-	if qb.translator.useNewResourceModel {
-		var filterBuilder *utils.FilterBuilder
-		if strings.HasPrefix(qb.metricName, "prometheus.googleapis.com") {
-			filterBuilder = utils.NewFilterBuilder(utils.SchemaTypes["prometheus"])
-		} else {
-			filterBuilder = utils.NewFilterBuilder(utils.SchemaTypes["pod"])
-		}
+	filterBulder := qb.GetFilterBuilder().
```
kwiesmueller, line 298:
```suggestion
	filterBuilder := qb.GetFilterBuilder().
```

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -19,12 +20,18 @@ type Schema struct {
 }
 
 const (
-	PodType        = "k8s_pod"
-	PrometheusType = "prometheus_target"
-	LegacyType     = "<not_allowed>"
+	PodSchemaKey        = "pod"               // PodSchemaKey is the key to use pod type filter schema.
```
kwiesmueller, line 23: Not sure if GoDoc picks these comments up correctly. If they are above the consts it will, but just a minor nit.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -89,100 +91,229 @@ type Translator struct {
...
+		WithProject(qb.translator.config.Project).
+		WithCluster(qb.translator.config.Cluster)
 
-		filter = filterBuilder.
-			WithMetricType(qb.metricName).
-			WithProject(qb.translator.config.Project).
-			WithCluster(qb.translator.config.Cluster).
+	if qb.translator.useNewResourceModel {
+		// new resource model specific filters
+		filterBulder.
```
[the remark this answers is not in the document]
bryan-hz: oh right, that's a good catch, thanks

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
bryan-hz: yep, GoDoc renders it, at least I use vscode, and it works for me. I think for variable type, put the doc after it instead of above it makes it a little bit neater, how do you think 🤔

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -89,100 +91,229 @@ type Translator struct {
...
+
+// Build is the last step for QueryBuilder which converts itself into a ProjectsTimeSeriesListCall object
+//   - has to pass the prerequisits specified in QueryBuilder.validate()
+//   - uses the query schema based on useNewResourceModel from translator as well as the metric name.
+//   - composes provided filters using the internal tool FilterBuilder
+//
+// Example:
+//
+//	projectsTimeSeriesListCall, error = NewQueryBuilder(translator, metricName).Build()
 func (qb *QueryBuilder) Build() (*stackdriver.ProjectsTimeSeriesListCall, error) {
```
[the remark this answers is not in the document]
bryan-hz: My understanding is that all this function does is to translate existing provided filters into a ProjectsTimeSeriesListCall. It doesn't destroy the original QueryBuilder object, you can build multiple times as long as the object it stored anywhere such as
```go
queryBuilder := NewQueryBuilder(translator, metricName)
q1 := queryBuilder.Build()
q2 := queryBuilder.Build()
```
not sure if this answers your question

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -0,0 +1,215 @@
...
+//
+// Example:
+//
+//  1. To add "resource.labels.pod_name = my-pod"
+//     filterBuilder := NewFilterBuilder(SchemaTypes[PodSchemaKey]).WithPods([]string{"my-pod"})
+//
+//  2. To add "resource.lables.pod_name = one_of(my-pod-1,my-pod-2)"
+//     filterBuilder := NewFilterBuilder(SchemaTypes[PodSchemaKey]).WithPods([]string{"my-pod-1", "my-pod-2"})
+func (fb *FilterBuilder) WithPods(pods []string) *FilterBuilder {
+	if len(pods) == 0 {
```
courageJ, line 195: It can be 

Option 1: Switch
```
switch len(pods) {
  case 0: 
      klog.Warningf("...")
  case 1:
      fb.filters = append(...)
  default:
      fb.filters = append(...)
}

return fb
```
Option 2: If and return
```
if len(pods) ==0 {
  klog....
  return fb
}
if len(pods) == 1 {
  fb.filters = ...
  return fbs
}
fb.filters = ...
return fbs
```

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder_test.go
```diff
@@ -0,0 +1,241 @@
+package utils
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+// Helper struct for testing syntax sugar
```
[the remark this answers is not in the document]
bryan-hz: oh I didn't know this, thanks for pointing it out!

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -88,98 +91,252 @@ type Translator struct {
...
+//   - metricName, required to determine the query schema.
+//
+// Example:
+//
+//	queryBuilder := NewQueryBuilder(NewTranslator(...), "custom.googleapis.com/foo")
+func NewQueryBuilder(translator *Translator, metricName string) *QueryBuilder {
+	return &QueryBuilder{
+		translator:           translator,
+		metricName:           metricName,
+		enforceContainerType: false,
```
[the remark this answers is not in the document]
bryan-hz: gotcha

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -88,98 +91,252 @@ type Translator struct {
...
+
+// WithMetricKind adds a metric kind filter to to the QueryBuilder
+//
+// Example:
+//
+//	queryBuilder := NewQueryBuilder(translator, metricName).WithMetricKind("GAUGE")
+func (qb *QueryBuilder) WithMetricKind(metricKind string) *QueryBuilder {
+	qb.metricKind = metricKind
+	return qb
+}
```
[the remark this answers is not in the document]
bryan-hz: right, this is to provide a [fluent interface](https://en.wikipedia.org/wiki/Fluent_interface) to support method chaining in the builder pattern.

for example, consider
```go
queryBuilder := NewQueryBuilder()
queryBuilder.setMetricKind(metricKind)
queryBuilder.setMetricValueType(metricValueType)
```
and 
```go
queryBuilder := NewQueryBuilder()
    .withMetricKind(metricKind)
    .withMetricValueType(metricValueType)
```
the latter solution gives better readability to me (and easier to write during development)

courageJ (approved):

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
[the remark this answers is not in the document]
kwiesmueller: Sgtm.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
kwiesmueller: Perfect, thx

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -184,6 +184,8 @@ func (fb *FilterBuilder) WithNamespace(namespace string) *FilterBuilder {
 
 // WithPods adds a filter for pods.
 //
+// # pods should NOT more than 100 according to "https://cloud.google.com/monitoring/api/v3/filters#comparisons"
```
kwiesmueller, line 187:
```suggestion
// Number of pods should NOT be more than 100 according to "https://cloud.google.com/monitoring/api/v3/filters#comparisons"
```

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -192,11 +194,18 @@ func (fb *FilterBuilder) WithNamespace(namespace string) *FilterBuilder {
 //  2. To add "resource.lables.pod_name = one_of(my-pod-1,my-pod-2)"
 //     filterBuilder := NewFilterBuilder(SchemaTypes[PodSchemaKey]).WithPods([]string{"my-pod-1", "my-pod-2"})
 func (fb *FilterBuilder) WithPods(pods []string) *FilterBuilder {
```
kwiesmueller, line 196: Is there a reason to have a pointer return type here? With the builder pattern it should not be needed afaik.

bryan-hz: gotcha, that's a good point, will update other places as well.

kwiesmueller (commented):
Thanks, lgtm.
Is there anything you can add to the existing tests (like client_test.go) to make sure the new code works as intended and the old code does too?

bryan-hz:
@kwiesmueller we doesn't add any new feature to core provider, the functionalities should be same as before, so I think we just need to make sure client_test.go still passes.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -232,16 +331,31 @@ func (qb QueryBuilder) validate() error {
 		return apierr.NewInternalError(fmt.Errorf("QueryBuilder tries to build with translator value: nil"))
 	}
 
-	if qb.pods != nil && qb.podNames != nil {
-		return apierr.NewInternalError(fmt.Errorf("both pods and podNames are set for QueryBuilder, expected only one of them should be used"))
+	if !qb.nodes.isNodesContainerEmpty() {
+		// node metric
+		if !qb.nodes.isNodesContainerValid() {
+			return apierr.NewInternalError(fmt.Errorf("invalid nodes parameter is set to QueryBuilder"))
```
kwiesmueller, line 337: These errors are a little implementation specific. I wonder if we can find errors that are more actionable to users.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/response_translator.go
```diff
@@ -243,6 +248,10 @@ func (t *Translator) metricKey(timeSeries *stackdriver.TimeSeries) (string, erro
 		case "k8s_node":
 			return ":" + timeSeries.Resource.Labels["node_name"], nil
 		case "prometheus_target":
+			if resourceSchema == "nodes" {
+				// ignore port number part
```
kwiesmueller, line 252: can you add an example for what this is about?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -54,6 +57,16 @@ var (
...
 		pods:         "resource.labels.pod_id",
 	}
+	// NodeSchema is the predefined schema for building node type queries.
+	NodeSchema = &Schema{
+		resourceType: "resource.type",
+		metricType:   "metric.type",
+		project:      "resource.labels.project_id",
+		cluster:      "resource.labels.cluster_name",
+		location:     "resource.labels.location",
+		namespace:    "resource.labels.namespace_name",
```
courageJ, line 67: Why do we have namespace for node?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -91,6 +91,74 @@ type Translator struct {
...
+		for _, item := range pc.pods.Items {
+			podNames = append(podNames, fmt.Sprintf("%q", item.GetName()))
+		}
+		return podNames
+	}
+	return pc.podNames
+}
+
+// getPodIds gets pod ids from podsContainer if any provided
+func (pc podsContainer) getPodIds() []string {
```
courageJ, line 123: Nit: ID vs Id https://github.com/golang/lint/issues/124

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder_test.go
```diff
@@ -779,7 +788,47 @@ func TestTranslator_GetSDReqForNodes(t *testing.T) {
 	}
 }
 
-func TestTranslator_GetSDReqForNodes_withMetricSelector(t *testing.T) {
+func TestQueryBuilder_Multiple_Node(t *testing.T) {
```
courageJ, line 791: Nit: multiple nodes

bryan-hz: yea, I double considered the naming as well, `Node` here I used singular refers to the type node, but definitely I'm good to change it to `nodes`

custom-metrics-stackdriver-adapter/pkg/adapter/translator/response_translator.go
bryan-hz: yep, for `k8s_node` type metrics, usually it's just the node id (`gke-stackdriver-adapter--default-pool-cee13989-0i75`), but for `prometheus_target` metrics, it will also include the port number who emits the metric (gke-stackdriver-adapter--default-pool-cee13989-0i75:8080).

I can add the explanation to the comment as well

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
bryan-hz: that's a good catch thanks

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -91,6 +91,74 @@ type Translator struct {
 	supportDistributions bool
 }
 
+// podsContainer is a helper struct to hold pods values
+type podsContainer struct {
```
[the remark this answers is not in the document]
bryan-hz: it's a container because it contains pod values, previously I named it as `podsHolder`, but I feel `podsContainer` is more descriptive, how do you think

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
bryan-hz: gotcha, thanks

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -91,6 +91,74 @@ type Translator struct {
...
+		for _, item := range pc.pods.Items {
+			podIds = append(podIds, fmt.Sprintf("%q", item.GetUID()))
+		}
+		return podIds
+	}
+	return []string{}
+}
+
+// nodesContainer is a helper struct to hold nodes values
+type nodesContainer struct {
```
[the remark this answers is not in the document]
bryan-hz: it holds two different ways of providing node values for query builder

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
bryan-hz: I thought about it, this case actually shouldn't happen, since all scenarios are covered by  `!qb.nodes.isNodesContainerEmpty()`

custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
```diff
@@ -37,13 +37,26 @@ subjects:
   namespace: custom-metrics
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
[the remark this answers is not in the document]
bryan-hz: sounds good to me

bryan-hz:
Hi @courageJ @kwiesmueller , resolved comments, please take a look when you get a chance, thanks 🙏

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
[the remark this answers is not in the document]
courageJ: Same as above, people will misunderstand it as k8s container.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder_test.go
courageJ: What is Multiple for here? Multiple filters or multiple nodes?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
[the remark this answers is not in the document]
bryan-hz: any suggestion on its naming (as well as for the node type)?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -276,10 +397,45 @@ func (qb QueryBuilder) getFilterBuilder() utils.FilterBuilder {
...
+		WithProject(qb.translator.config.Project).
+		WithCluster(qb.translator.config.Cluster)
+
+	resourceNames := qb.getResourceNames()
+	if qb.translator.useNewResourceModel {
+		// new resource model specific filters
+		filterBuilder = filterBuilder.WithLocation(qb.translator.config.Location)
+		if !qb.nodes.isNodesContainerEmpty() {
+			// node metrics
+			filterBuilder = filterBuilder.WithNodes(resourceNames)
```
courageJ, line 423: Can we just return 
```
filterBuilder = filterBuilder.WithNodes(resourceNames)
return filterBuilder.Build()
```
to reduce the usage of if else?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/response_translator.go
```diff
@@ -243,6 +248,11 @@ func (t *Translator) metricKey(timeSeries *stackdriver.TimeSeries) (string, erro
 		case "k8s_node":
 			return ":" + timeSeries.Resource.Labels["node_name"], nil
 		case "prometheus_target":
+			if resourceSchema == "nodes" {
+				// for prometheus_target metric, its instance usually consist of node id and the port who emited the metric.
```
courageJ, line 252: I think we want to use the node label from the metrics instead of the instance. For example, a kube-state-metrics metrics exposes `kube_node_status_condition` (https://github.com/kubernetes/kube-state-metrics/blob/master/docs/node-metrics.md), the instance will be the node that KSM is deployed. But we want to use the node label exported from the metric.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
bryan-hz: sure

custom-metrics-stackdriver-adapter/pkg/adapter/translator/response_translator.go
bryan-hz: an example prometheus metric could be like this
```json
{ 
     "metric": {
        "labels": {
          "pod": "node-exporter-lfx95"
        },
        "type": "prometheus.googleapis.com/node_load1/gauge"
      },
      "resource": {
        "type": "prometheus_target",
        "labels": {
          "location": "",
          "project_id": "",
          "job": "node-exporter",
          "instance": "gke-stackdriver-adapter--default-pool-cee13989-0i75:8080",
          "cluster": "stackdriver-adapter-dev",
          "namespace": ""
        }
}
```
there is not always a label called `node`, or I misunderstood

courageJ: for some metrics, there will be a node label, like `kube_node_status_condition` (You can follow https://cloud.google.com/stackdriver/docs/managed-prometheus/exporters/kube_state_metrics to try kube-state-metrics metrics). For some metrics from node-exporter, as your example, the node name can be inferred from the instance. 

Can we do something like this, if there is a node metric label, we use the node value; otherwise, we use the instance value?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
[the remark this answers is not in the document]
bryan-hz: in this case, how do you think of `podValues`? (since it holds two types of pod/nod values)

custom-metrics-stackdriver-adapter/pkg/adapter/translator/response_translator.go
```diff
@@ -249,6 +249,10 @@ func (t *Translator) metricKey(timeSeries *stackdriver.TimeSeries, resourceSchem
 			return ":" + timeSeries.Resource.Labels["node_name"], nil
 		case "prometheus_target":
 			if resourceSchema == "nodes" {
+				// use node label if exists
+				if nodeName, ok := timeSeries.Resource.Labels["node"]; ok {
```
courageJ, line 253: It should be timeSeries.Metric.Labels.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -91,6 +91,80 @@ type Translator struct {
...
+// for example, when WithPods() and WithPodNames() are both used in QueryBuilder
+func (pc podValues) isPodsContainerValid() bool {
+	return !(len(pc.podNames) > 0 && pc.pods != nil && len(pc.pods.Items) > 0)
+}
+
+// isPodsEmpty checks if pods container is empty
+func (pc podValues) isPodsContainerEmpty() bool {
+	return len(pc.podNames) == 0 && (pc.pods == nil || len(pc.pods.Items) == 0)
+}
+
```
courageJ, line 112: Should we also update the function name since the struct name is updated?

courageJ: If you want to comment a function, please keep the name consistency. It uses `isPodsEmpty` in comment but the function name is isPodsContainerEmpty.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -259,6 +379,7 @@ func (qb QueryBuilder) validate() error {
 //  1. if useNewResourceModel is false, then use legacy type FilterBuiler
 //  2. if enforceContainerType is true, then use container type FilterBuilder
 //  3. if metricName is prefixed with PrometheusMetricPrefix, then use prometheus type FilterBuilder
+//  4. if namespace is empty, then use node type FilterBuilder
 //  4. By default, use pod type FilterBuilder
```
courageJ, line 383: Nit: 5?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -62,13 +74,15 @@ var (
 		cluster:      "resource.labels.cluster",
 		location:     "resource.labels.location",
 		namespace:    "resource.labels.namespace",
+		nodes:        "resource.labels.instance",
```
courageJ, line 77: Do we also want to update it here?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -212,6 +228,33 @@ func (fb FilterBuilder) WithPods(pods []string) FilterBuilder {
 	return fb
 }
 
+// WithNodes adds a filter for nodes.
```
courageJ, line 231: Do we want to update it here?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder_test.go
```diff
@@ -245,6 +245,52 @@ func TestFilterBuilder_WithPods_Multiple_legacy(t *testing.T) {
 	expectFilterBuilder(actual).toEqual(expected).report(t)
 }
```
courageJ, line 247: Please add test cases for node metric labels.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
bryan-hz: good catch

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
bryan-hz: I did some testing, I think you are right, "metric.labels.node" is the correct way to get the node name: 
- if we filter by "resource.labels.instance", then we will get timeseries that contains both value with "node" label and value without "node" label, which doubles the result.
- if we filter by "metric.labels.node", then we will only the the value with "node" label.
i will update all the use cases, thanks for pointing it out

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder_test.go
bryan-hz: as discussed, we are only using `metric.labels.node` now, so it's covered by other tests

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
bryan-hz: no, we don't need to update this method

courageJ (approved):

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -91,64 +91,70 @@ type Translator struct {
...
 
-// getPodNames gets pod names from podsContainer is any provided
-func (pc podsContainer) getPodNames() []string {
-	if pc.pods != nil {
-		podNames := []string{}
-		for _, item := range pc.pods.Items {
-			podNames = append(podNames, fmt.Sprintf("%q", item.GetName()))
-		}
-		return podNames
+// getPodNames gets pod names from podValues is any provided
```
kwiesmueller, line 113:
```suggestion
// getPodNames gets pod names from podValues if any provided
```

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -91,64 +91,70 @@ type Translator struct {
...
+func (pc podValues) getPodNames() []string {
+	if pc.pods == nil {
+		return pc.podNames
 	}
-	return pc.podNames
+	podNames := make([]string, len(pc.pods.Items))
+	for i, item := range pc.pods.Items {
+		podNames[i] = fmt.Sprintf("%q", item.GetName())
```
kwiesmueller, line 120: Why do you need to quote the names?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -91,64 +91,70 @@ type Translator struct {
...
+func (nc nodeValues) isNodeValuesEmpty() bool {
 	return len(nc.nodeNames) == 0 && (nc.nodes == nil || len(nc.nodes.Items) == 0)
 }
 
 // getNodeNames gets node names from nodeContainer if any provided
-func (nc nodesContainer) getNodeNames() []string {
+func (nc nodeValues) getNodeNames() []string {
 	if nc.nodes != nil {
 		nodeNames := []string{}
 		for _, item := range nc.nodes.Items {
```
kwiesmueller, line 160: And why are you not quoting the node names too?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -231,27 +230,20 @@ func (fb FilterBuilder) WithPods(pods []string) FilterBuilder {
...
-//	need to provide target node names (such as "gke-test--default-pool-cee13989-qsky"), because we
-//	have "(:\\d+)*" at the end of the regex to ignore the port number part.
-//
 // Example:
 //
-//	// To filter for instances gke-test--default-pool-cee13989-0i75 and
-//	// gke-test--default-pool-cee13989-qsky:8080 with the query
-//	// resource.labels.instance = monitoring.regex.full_match("^" +
-//	// 	"(gke-test--default-pool-cee13989-0i75|gke-test--default-pool-cee13989-qsky)" +
-//	// 	"(:\\d+)*"
```
kwiesmueller: Can you explain this change in query syntax?

bryan-hz: originally I was trying to filter something in format `node-name:5000` for resource.labels.instance,
1. "^", first part marks start of node name
2. "(node1|node2)", second part is saying it should be either `node1` or `node2`
3. "(:\\d+)*)", third part is saying it's possible to followed by a port number `:5000` (0 or mulultiple), but it cannot be only `:`
since we changed to use metric.labels.node, it will not contain `:5000`, we don't need the third part any more, and in later commits it was changed to `$` which marks the end of name.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
bryan-hz: actually I think I misunderstood your question, pod names are quoted and node names are not quoted, that's because they have different usages for node names and pod names.

for pod names: `one_of("pod1", "pod2")`
for node names: `monitoring.regex.full_match("^(node1|node2)$")`

for regex method, it doesn't have the filtering number constraint, but it's less readable and hard to maintain, so I didn't change the existing usage implementation for filtering pods. If we encounter some usecases that requires to remove the 100 pod filtering constraint in `one_of`, then we can switch to use regex method.

kwiesmueller: Okay makes sense, can you clarify that in the function name? Like getQuotedPodNames?
Or do the quoting when you add them to the query. Either way works for me, whatever you prefer.

bryan-hz: sure

kwiesmueller (approved):
Thanks!

osalau (approved):

PR #491, opened 2022-11-09
[Enhance] Update autoscaling api version to v2 in prometheus example

bryan-hz:
Provided that `autoscaling/v2beta2` will be deprecated since `Kubernetes v1.26`, and `autoscaling/v2` version has all the features needed in the example. We change the HPA api versions to `autoscaling/v2` in examples.

According to the Kubernetes [doc](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#scaling-on-custom-metrics), scaling on custom metrics feature after _Kubernetes v1.23 Stable_. Therefore, from my perspective, updating the HPA api version in the example to use `autoscaling/v2` should be sufficient.


bryan-hz:
Hi @courageJ @laoj2, please help take a look when you get a chance.

laoj2 (approved):

PR #493, opened 2022-11-16
Fix prometheus example

bryan-hz:
Verified with
```sh
docker run -p 8080:8080 gcr.io/google-samples/prometheus-dummy-exporter:v0.3.0 "./prometheus-dummy-exporter" "--metric-value=40" "--metric-name=foo" "--port=8080"
```


bryan-hz:
Hi @courageJ @laoj2, please help take a look when you get a chance.

custom-metrics-stackdriver-adapter/examples/prometheus-to-sd/Dockerfile
```diff
@@ -12,10 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.12.5
-ADD . /go/src/prometheus-dummy-exporter
-RUN go get github.com/prometheus/client_golang/prometheus
+FROM golang:1.16-alpine AS builder
```
courageJ, line 15: why "golang:1.16-alpine" not "golang:1.16"?

custom-metrics-stackdriver-adapter/examples/prometheus-to-sd/go.mod
```diff
@@ -0,0 +1,17 @@
+module prometheus-dummy-exporter
+
+go 1.20
```
courageJ, line 3: Why it is 1.20 here while it is go1.16 in Dockerfile?

custom-metrics-stackdriver-adapter/examples/prometheus-to-sd/Dockerfile
```diff
@@ -12,10 +12,18 @@
...
+
+WORKDIR /app
+
+COPY go.mod ./
+COPY go.sum ./
+RUN go mod download
```
courageJ, line 21: do we want to have a vendor folder instead of running go mod download when building?

custom-metrics-stackdriver-adapter/examples/prometheus-to-sd/Dockerfile
laoj2, line 15: why aren't you using 1.19-alpine here?

bryan-hz: right, `alphine` is a more secure and lightweight version of linux

custom-metrics-stackdriver-adapter/examples/prometheus-to-sd/go.mod
bryan-hz: good catch

custom-metrics-stackdriver-adapter/examples/prometheus-to-sd/Dockerfile
bryan-hz: this is a intermediate step which is going to be discarded anyways, and the base image in the last stage `gcr.io/distroless/static` makes sure our final image only contains an executable file

courageJ (approved):

custom-metrics-stackdriver-adapter/Makefile
```diff
@@ -3,7 +3,7 @@ GOOS?=linux
 OUT_DIR?=build
 PACKAGE=github.com/GoogleCloudPlatform/k8s-stackdriver/custom-metrics-stackdriver-adapter
 PREFIX?=staging-k8s.gcr.io
-TAG = v0.12.2
+TAG = v0.13.0
```
laoj2, line 6: we're just changing the examples, why are you bumping the adapter's version?

bryan-hz: yesterday when we do the release, I notice that the image created was tagged to `v0.12.2` instead of `v0.13.0`, I can create a separate PR for this

laoj2: Sounds good → let's do it in a separate PR, since this is not related to this PR's fix.

laoj2 (approved):

PR #496, opened 2022-11-21
fix: remove namespaces from the clusterrole

### LINKED ISSUE #492: ClusterRole in stackdriver adapter has namespace defined
In file [adapter_new_resource_model.yaml on line 44](https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml#L42) `namespace` property is defined in the `metadata` of the **ClusterRole**.

From the official [documentation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole):

> ClusterRole, by contrast, is a non-namespaced resource. The resources have different names (Role and ClusterRole) because a Kubernetes object always has to be either namespaced or not namespaced; it can't be both.

I noticed this problem because my terraform apply was failing with the logs:
```
Error: Cluster level resource cannot take namespace
with kubernetes_manifest.stackdriver["ClusterRole/custom-metrics-resource-reader"]
Resources of type 'rbac.authorization.k8s.io/v1, Kind=ClusterRole' cannot have a namespace
```
[Here](https://github.com/GoogleCloudPlatform/k8s-stackdriver/commit/d0e6643438e2ea83bc448603668ca1dd7b7e0edf#diff-0050544198601f6f008f5f100331d61f5f8a28da2725380ca519f3ff8142851dR43) is the relevant commit when namespace was added to the CLusterRole

ajayk:
fixes #492

### DIFF
--- a/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
+++ b/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
@@ -40,7 +40,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: custom-metrics-resource-reader
-  namespace: custom-metrics
 rules:
 - apiGroups:
   - ""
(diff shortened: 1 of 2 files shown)


juli4n:
Thanks for the contribution. This was fixed already in https://github.com/GoogleCloudPlatform/k8s-stackdriver/pull/941.

PR #533, opened 2023-04-04
fix: fixes invisible args in custom-metrics-stackdriver-adapter README

ragrag:
fixes invisible args in custom-metrics-stackdriver-adapter README


CatherineF-dev:
Thx!

PR #552, opened 2023-09-22
Install Dependabot to support Docker and go module updates

wespanther:
Enable Dependabot for `Dockerfile` and `go.mod` dependency updates. 
This should hopefully help keep our go module and base / builder images up to date.

The PR was generated by a script, I used this as an opportunity to do a sanity check. Let me know if I should change any of the configuration values.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: docker
+  directory: /
+  schedule:
+    interval: daily
+- package-ecosystem: gomod
+  directory: /
+  schedule:
+    interval: daily


wespanther:
Also, let me know if we should change the frequency - does "weekly" sound more tenable than "daily"?

kwiesmueller:
Daily should be fine for now, but I wonder if the directory rules will correctly detect all mod and docker files in subdirectories. So far it seemed to be working.

kwiesmueller (approved):

wespanther:
Thanks @kwiesmueller!

> Daily should be fine for now, but I wonder if the directory rules will correctly detect all mod and docker files in subdirectories.

It should detect all of the subdirectories, but it will also include event-adapter in `/archived/`, which me may or may not want. We can ignore (or merge) those PRs, or we can explicitly add each directory in the config (but we'd lose out on any new ones that are added).

If you're good with this, can you merge it? I don't have write on this repo.

PR #554, opened 2023-10-03
Update dependabot config to explicitly list directories

wespanther:
This should make ensure dependabot operates on each, and ignore the "archived" directory.


wespanther:
@erain - I don't have permission to merge, can you do it for me? Thank you!

Issue #485, opened 2022-09-12
Custom stack driver pod doesnt seem to have liveness/readiness probe.


dmjdell:
Currently using custom stack driver and there is a requirement to add liveness/readiness probe for custom stack driver. Can we add this feature ?

PR #710, opened 2024-05-06
Add readiness and liveness probe

CatherineF-dev:
Fixes: [#485](https://github.com/GoogleCloudPlatform/k8s-stackdriver/issues/485)

Tested: 
```
curl -k https://localhost:6443/livez
ok

curl -k https://localhost:6443/readyz
ok

k top pods -n custom-metrics
NAME                                                 CPU(cores)   MEMORY(bytes)   
custom-metrics-stackdriver-adapter-cddcdcf54-vdhlj   3m           32Mi 
```

### DIFF
--- a/custom-metrics-stackdriver-adapter/deploy/production/adapter.yaml
+++ b/custom-metrics-stackdriver-adapter/deploy/production/adapter.yaml
@@ -78,6 +78,23 @@ spec:
         command:
         - /adapter
         - --use-new-resource-model=false
+        - --secure-port=6443
+        readinessProbe:
+          failureThreshold: 5
+          httpGet:
+            path: /readyz
+            port: https
+            scheme: HTTPS
+          initialDelaySeconds: 30
+          periodSeconds: 5
+        livenessProbe:
+          failureThreshold: 5
+          httpGet:
+            path: /livez
+            port: https
+            scheme: HTTPS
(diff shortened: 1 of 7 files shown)


CatherineF-dev:
The failed test isn't related to this change. 

The failed one is a flaky test, in checking

custom-metrics-stackdriver-adapter/deploy/production/adapter.yaml
```diff
@@ -78,6 +78,26 @@ spec:
         command:
         - /adapter
         - --use-new-resource-model=false
+        - --secure-port=6443
+        ports:
+        - containerPort: 6443
+          name: https
```
leonzz, line 84: is this port only for serving the readiness and liveness checks?  if so consider giving it a more descriptive name or just use port number on line 89 and 97 without a name.
If it is also serving the metrics API, then I guess we also need to update the `targetPort: 443` for the Service?

CatherineF-dev: Good point! Nice catch!
Updated to `targetPort: 6443` and did an E2E test following https://cloud.google.com/stackdriver/docs/managed-prometheus/hpa#stackdriver-adapter

leonzz (approved):

PR #739, opened 2024-07-13
prom-to-sd: Prefer proto in content type

linxiulei:
Also add support for multiple response formats (text/proto)

### DIFF
--- a/prometheus-to-sd/translator/translator.go
+++ b/prometheus-to-sd/translator/translator.go
@@ -358,11 +358,13 @@ func convertToDistributionValue(h *dto.Histogram) *v3.Distribution {
 
 	prevVal := uint64(0)
 	lower := float64(0)
+	infSeen := false
 	for _, b := range h.Bucket {
 		upper := b.GetUpperBound()
 		if !math.IsInf(b.GetUpperBound(), 1) {
 			bounds = append(bounds, b.GetUpperBound())
 		} else {
+			infSeen = true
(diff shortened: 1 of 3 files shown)


CatherineF-dev:
cc @leonzz to have a review as well

prometheus-to-sd/translator/translator.go
```diff
@@ -375,6 +377,11 @@ func convertToDistributionValue(h *dto.Histogram) *v3.Distribution {
 		prevVal = b.GetCumulativeCount()
 	}
 
+	// +Inf Bucket is implicit so it needs to be added
+	if !infSeen && count > int64(prevVal) {
```
leonzz, line 381: looks like this is addressing a different issue, could you add some background in the PR description?

linxiulei: it's not a different issue because the +inf bucket is omitted in protobuf encoding but text encoding handles that specially (see https://github.com/prometheus/common/blob/main/expfmt/text_create.go#L232 if interested). However, it makes sense to separate the changes with description, for which I split the changes into 2 commits to make it clear.

leonzz: Thanks!  LGTM-)

leonzz (approved):

CatherineF-dev:
Will merge. Have discussed offline that the release flow will be expected.

PR #742, opened 2024-07-17
Replace golang.org/api/monitoring/v3 with cloud.google.com/go/monitoring/apiv3 

linxiulei:
As `google.golang.org/api/monitoring/v3` is deprecated with suboptimal performance and whatnot, it's recommended to use package `cloud.google.com/go/monitoring/apiv3` instead.

An additional note: the most outstanding improvement in performance is that cloud.google.com/go/monitoring/apiv3 uses protobuf instead of json, which is used in google.golang.org/api/monitoring/v3. That decreases the resource overheads of data encodings when posting metric to cloud monitoring service.


leonzz (approved):
looks good to me. Thanks @linxiulei!

prometheus-to-sd/main.go
```diff
@@ -144,32 +143,18 @@ func main() {
...
-		}
-		glog.Infof("Created a client with the default credentials")
-	} else {
-		ts, err := google.DefaultTokenSource(context.Background(), "https://www.googleapis.com/auth/cloud-platform")
-		if err != nil {
-			glog.Fatalf("Error creating default token source: %v", err)
-		}
-		client = oauth2.NewClient(context.Background(), ts)
+		ts := config.NewAltTokenSource(*gceTokenURL, *gceTokenBody)
+		options = append(options, option.WithTokenSource(ts))
```
CatherineF-dev, line 149: > *projectOverride != ""
I find this else is deleted. Will it respect projectOverride?

prometheus-to-sd/main.go
```diff
@@ -144,32 +143,18 @@ func main() {
 		glog.Error(http.ListenAndServe(fmt.Sprintf("%s:%d", *debugAddress, *debugPort), expvar.Handler()))
 	}()
 
-	var client *http.Client
-
+	var options []option.ClientOption
 	if *gceTokenURL != "" {
```
CatherineF-dev, line 147: Do we need to handle `else` case?

prometheus-to-sd/main.go
linxiulei: Yes, the project name is in the request body,  see https://github.com/GoogleCloudPlatform/k8s-stackdriver/pull/742/files#diff-fabb01b5a0dd10f9e48b220cf31c93d51c6fd5a5516b4254ca63232c43a0dfe8R61

prometheus-to-sd/main.go
linxiulei: First, two `else`s before this PR were actually doing the same thing (ie. `google.DefaultClient()` creates google.DefaultTokenSource() and puts it in oauth2.NewClient)
Second, the projectOverride is not needed because everytime the client posts metrics, the request body contains the project name (see https://github.com/GoogleCloudPlatform/k8s-stackdriver/pull/742/files#diff-fabb01b5a0dd10f9e48b220cf31c93d51c6fd5a5516b4254ca63232c43a0dfe8R61)

prometheus-to-sd/main.go
leonzz: +1 to @linxiulei's comment. if you trace what's being done by [google.DefaultClient()](https://cs.opensource.google/go/x/oauth2/+/ee480838109b20d468babcb00b7027c82f962065:google/default.go;l=73) in the `*projectOverride != ""` case, it's exactly the same as the other case. The projectOverride string is not used at all.

PR #743, opened 2024-07-18
Add a fake metric to the external metrics listing api

bryan-hz:
`Helm install` will fail with empty metrics list, this PR should fix the issue.


jacobycwang:
Sweet, thanks @bryan-hz a lot, besides, do we already have a schedule on when it will be release?

CatherineF-dev:
Hi @jacobycwang, eta is 2 weeks (release a new image at gcr.io and update image version in this repo).

PR #784, opened 2024-10-17
prom-to-sd: Fix not respecting --api-override

linxiulei:
/cc @CatherineF-dev @leonzz

### DIFF
--- a/prometheus-to-sd/main.go
+++ b/prometheus-to-sd/main.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


prometheus-to-sd/main.go
```diff
@@ -259,3 +269,32 @@ func parseMonitoredResourceLabels(monitoredResourceLabelsStr string) map[string]
 	}
 	return labels
 }
+
+func getEndpoint(input string) (string, error) {
```
CatherineF-dev, line 273: add unit test

leonzz (approved):

prometheus-to-sd/main.go
linxiulei: where do you suggest we put these functions to? we can't have `main_test.go`, can we?

leonzz: I think it's ok to not have unit test for this method because:
1. we've manually verified it with https://go.dev/play/p/pMP9_m2ZWXK
2. this is during startup so future issues in URL parsing will be easily catched by e2e test

CatherineF-dev: Ok.

PR #789, opened 2024-11-05
Fix the fallback for container metrics logic to query both container and pod metrics

nabil-dbz:
The main idea here is to avoid failure when querying pod level metrics throw an error. For example for the metric type is `kubernetes.io/container/accelerator/duty_cycle`, this query:
```
metric.type = "kubernetes.io/container/accelerator/duty_cycle" AND resource.type = "k8s_pod"
```

Throws the following error:
```
The supplied filter does not specify a valid combination of metric and monitored resource descriptors. 
The query will not return any time series.
```

Which makes the adapter return an error before trying to query the metrics using the `k8s_container`.

The proposed solution for this is to add a new resource type called `PodContainerType` for which we use the operator `one_of` to handle the fallback logic. However, given that the response might contain both `k8s_pod` and `k8s_container` metrics (time series), we're adding a post-processing step to consider `k8s_container` metrics if `k8s_pod` metrics are absent.

Also, to support resource label filters for custom metrics, we're adding `resource.labels` to the list of allowed prefixes for custom metrics.

### DIFF
--- a/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
+++ b/custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
@@ -188,6 +188,9 @@ func (p *StackdriverProvider) getNamespacedMetricByName(groupResource schema.Gro
 		return nil, err
 	}
 	queryBuilder := translator.NewQueryBuilder(p.translator, metricName)
+	if p.fallbackForContainerMetrics {
+		queryBuilder = queryBuilder.EnforcePodContainerType()
+	}
 
 	pods := &v1.PodList{Items: []v1.Pod{*matchingPod}}
 	stackdriverRequest, err := queryBuilder.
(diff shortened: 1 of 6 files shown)


raywainman:
I will take a close look at this too, adding this to my list for this week. Thanks @nabil-dbz!

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -349,6 +311,27 @@ func (p *StackdriverProvider) ListAllExternalMetrics() []provider.ExternalMetric
 	}
 }
 
+func (p *StackdriverProvider) PostProcessPodContainerResp(response *stackdriver.ListTimeSeriesResponse, metricName string) (*stackdriver.ListTimeSeriesResponse, error) {
```
raywainman, line 314: We should really document what this does because modifying a response object here could be really confusing.

I was initially going to suggest not returning a ListTimeSeriesResponse here but I see that this would cause a bunch of issues with the follow-up calls.

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -41,7 +41,7 @@ var (
 	allowedExternalMetricsFullLabelNames = []string{"resource.type", "reducer"}
 	// allowedCustomMetricsLabelPrefixes and allowedCustomMetricsFullLabelNames specify all metric labels allowed for querying
 	allowedCustomMetricsLabelPrefixes  = []string{"metric.labels"}
-	allowedCustomMetricsFullLabelNames = []string{"reducer"}
+	allowedCustomMetricsFullLabelNames = []string{"resource.labels.container_name", "reducer"}
```
raywainman, line 44: I think you can omit `resource.labels` here, can't you? since it will be picked up by the prefix above?

custom-metrics-stackdriver-adapter/pkg/adapter/translator/utils/filter_builder.go
```diff
@@ -107,6 +113,8 @@ func NewFilterBuilder(resourceType string) FilterBuilder {
 	switch resourceType {
 	case PodType:
 		schema = PodSchema
+	case PodContainerType:
```
raywainman, line 116: What if we named this something like `PodOrContainerType`?

Otherwise this gets confusing :(

PR #837, opened 2025-03-07
Add a cache for custom metrics adapter to store response from Cloud Monitoring

richabanker:


custom-metrics-stackdriver-adapter/adapter.go
```diff
@@ -211,6 +213,8 @@ func main() {
 		"Stackdriver Endpoint used by adapter. Default is https://monitoring.googleapis.com/")
 	flags.BoolVar(&serverOptions.EnableDistributionSupport, "enable-distribution-support", serverOptions.EnableDistributionSupport,
 		"enables support for scaling based on distribution values")
+	flags.DurationVar(&serverOptions.ExternalMetricCacheWindow, "external-metric-cache-window", serverOptions.ExternalMetricCacheWindow,
```
raywainman, line 216: for additional safety, should we add another boolean flag here to control the enablement of the cache?

Most users will want that off so we can completely bypass creating the cache, etc...

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -316,6 +318,17 @@ func (p *StackdriverProvider) ListAllMetrics() []provider.CustomMetricInfo {
...
 // GetExternalMetric queries Stackdriver for external metrics.
 func (p *StackdriverProvider) GetExternalMetric(ctx context.Context, namespace string, metricSelector labels.Selector, info provider.ExternalMetricInfo) (*external_metrics.ExternalMetricValueList, error) {
+	key := cacheKey{
+		namespace:      namespace,
+		metricSelector: metricSelector,
+		info:           info,
+	}
+	resp, ok := p.externalMetricCache.Get(key)
+
+	if ok {
```
raywainman, line 328: for additional telemetry, what do you think about adding some metrics to monitor how times we are getting cache hits?

(We can definitely do this in a future PR)

raywainman:
cc @CatherineF-dev

custom-metrics-stackdriver-adapter/pkg/adapter/provider/cache.go
```diff
@@ -0,0 +1,77 @@
...
+	"sync"
+	"time"
+
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/metrics/pkg/apis/external_metrics"
+	"sigs.k8s.io/custom-metrics-apiserver/pkg/provider"
+)
+
+// Cache for GetExternalMetric responses.
+type externalMetricCache struct {
```
CatherineF-dev, line 29: Should we add a cache size?  The cache size is growing with more HPA requests and might be OOM.

raywainman: Ah good point.. We may want to eventually do some garbage collection, otherwise the cache may grow really big.

For most clusters with a fairly static set of HPAs, the number of entries should stay fairly stable but in clusters with a lot of churn, I could imagine this getting very big.

So perhaps we need a separate thread that is periodically cleaning up the cache?

custom-metrics-stackdriver-adapter/adapter.go
richabanker: So if externalMetricCacheWindow == 0, the cache will not be constructed and we effectively fall back to the previous flow. Does that suffice or do we still need an additional flag?

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
richabanker: Sounds good, can track this in a separate PR. Added a TODO for it (in the Get() method).

custom-metrics-stackdriver-adapter/pkg/adapter/provider/cache.go
raywainman: For v1, I think I'd prefer as simple of a cache as possible. The OOM problem can be solved with giving the addon-resizer extra memory. I haven't verified but I believe the cache entries here should be really small, even with hundreds of HPAs, the memory footprint should be quite small.

custom-metrics-stackdriver-adapter/adapter.go
raywainman: I think that works!

custom-metrics-stackdriver-adapter/pkg/adapter/provider/cache.go
raywainman: The LRUExpireCache looks great here, I think we just need to be careful with it because it isn't threadsafe but looks like it can be an almost straight drop-in replacement for the cache here :) Good find!

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -44,6 +45,8 @@ import (
 const (
 	nodeResource = "nodes"
 	podResource  = "pods"
+	// Maximum number of entries in the external metrics LRU cache.
+	defaultExternalMetricsCacheSize = 1024
```
richabanker, line 49: @raywainman @CatherineF-dev do you think we need to expose this also as a flag to be configurable by users?

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
```diff
@@ -334,9 +370,21 @@ func (p *StackdriverProvider) GetExternalMetric(ctx context.Context, namespace s
...
-	}, nil
+	}
+
+	if p.externalMetricCache != nil {
+		key := cacheKey{
+			namespace:      namespace,
+			metricSelector: metricSelector.String(),
+			info:           info,
+		}
+		p.externalMetricCache.Add(key, resp, p.externalMetricsCacheTTL)
```
raywainman, line 384: let's also make sure to use a mutex here as I don't believe this cache is thread safe

it is possible for HPA to be ran in parallel which could cause us to write to this from multiple threads at the same time

richabanker: I think its thread-safe actually https://github.com/kubernetes/apimachinery/blob/v0.32.3/pkg/util/cache/lruexpirecache.go#L106-L107 ?

raywainman: ah perfect, thanks!

custom-metrics-stackdriver-adapter/pkg/adapter/provider/provider.go
richabanker: Removed the default values and using whatever values are passed to flags for now. This is because, we want to provide the cache as an optional configuration - without any defaults. Can revisit this later once we enable the cache by default.

custom-metrics-stackdriver-adapter/adapter.go
```diff
@@ -211,6 +215,10 @@ func main() {
 		"Stackdriver Endpoint used by adapter. Default is https://monitoring.googleapis.com/")
 	flags.BoolVar(&serverOptions.EnableDistributionSupport, "enable-distribution-support", serverOptions.EnableDistributionSupport,
 		"enables support for scaling based on distribution values")
+	flags.DurationVar(&serverOptions.ExternalMetricCacheTTL, "external-metric-cache-ttl", serverOptions.ExternalMetricCacheTTL,
+		"The duration (e.g., 1m, 5s) for which external metric values are cached.")
+	flags.IntVar(&serverOptions.ExternalMetricCacheSize, "external-metric-cache-size", serverOptions.ExternalMetricCacheSize,
```
raywainman, line 220: Could we set a default value here? I think 300 items here is more than enough.

This would effectively represent 300 HPAs in the cluster all using external metrics which is plenty, and folks who need it can boost this higher.

raywainman (commented):
/lgtm

custom-metrics-stackdriver-adapter/adapter.go
richabanker: Done

PR #845, opened 2025-03-13
prometheus-to-sd: add universe domain command flag

shuaich:
## Context

Stackdriver runs in multiple universe besides the default universe. It requires specifying the universe domain explicitly in the client when accessing monitoring service in Stackdriver.

This PR adds a new command line flag `--universe-domain`, when not specified, prometheus-to-sd infers universe domain from api endpoint.


prometheus-to-sd/main.go
```diff
@@ -150,13 +151,30 @@ func main() {
...
+			uDomain = strings.TrimSuffix(domain, "/")
+		}
+	}
+	// override inferred universe domain with specified value
+	if *universeDomain != "" {
+		uDomain = *universeDomain
+	}
+	if uDomain != "" {
+		glog.Infof("Universe domain is %s", uDomain)
+		options = append(options, option.WithUniverseDomain(uDomain))
```
CatherineF-dev, line 177: nit: only add WithUniverseDomain when universeDomain is not ""

shuaich: uDomain is an variable that can be overridden by the command line flag universeDomain.

CatherineF-dev: Why does it need to add the option for the default universe?

shuaich: It's default universe when it's not specified in the client options.

CatherineF-dev: Why was it working without this option?

shuaich: Discussed offline, agreed on the approach.

shuaich: Thank you for the prompt review.

PR #856, opened 2025-04-01
event-exporter: pod owner label for GCP cloud logging

leonzz:
Add logic for event exporter to get pod and infer the top level controller type and top level controller name. The new logic runs when an event is in k8s_pod monitored resource type and not in one of the system namespaces.
    
Also bump k8s and x/net dependencies to latest.

The dependency bumps have huge diff and hard to review, so they are put into a separate commit, to make the logic changes easier to review.


leonzz:
+ @baizhenyu @CatherineF-dev to review

event-exporter/kubernetes/watchers/events/handler.go
```diff
@@ -42,7 +42,7 @@ func newEventHandlerWrapper(handler EventHandler) *eventHandlerWrapper {
 	}
 }
 
-func (c *eventHandlerWrapper) OnAdd(obj interface{}) {
+func (c *eventHandlerWrapper) OnAdd(obj interface{}, isInInitialList bool) {
```
baizhenyu, line 45: What is `isInInitialList ` used for?

baizhenyu: Got it, thanks

event-exporter/kubernetes/podlabels/cache_collector.go
```diff
@@ -0,0 +1,159 @@
...
+	return pc, nil
+}
+
+// Get returns the label mappings for a given Pod. If there's no local cache, we will attempt a
+// get from apiserver.
+func (pc *podLabelCollectorWithCache) GetLabels(namespaceName, podName string) map[string]string {
+	if _, ok := pc.ignoredNamespaces[namespaceName]; ok {
+		return nil
+	}
+	if labels, ok := pc.cache.Get(cacheKey{Namespace: namespaceName, Name: podName}); ok {
```
baizhenyu, line 72: I noticed the recreation of `cacheKey{Namespace: namespaceName, Name: podName}` struct many times in this function, maybe we can define it as a variable?

baizhenyu (approved):

event-exporter/kubernetes/podlabels/cache_collector.go
leonzz: Done.

CatherineF-dev:
Should we update tag here? https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/event-exporter/Makefile#L23

event-exporter/kubernetes/podlabels/cache_collector.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+		} else {
+			recordPodGet("UnknownFailure")
+		}
+		return nil
+	}
+	recordPodGet("OK")
+	labels := getLabelsFromPod(pod)
+	if len(labels) > 0 {
+		pc.cache.Add(cacheKey{Namespace: namespaceName, Name: podName}, labels)
```
CatherineF-dev: Should we call pc.emptyLabelPodCache.Add(cacheKey{Namespace: namespaceName, Name: podName}, time.Now()) to refresh TTL here?

leonzz:
> Should we update tag here? https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/event-exporter/Makefile#L23

good catch.  I'll do the necessary change and cut new version tag in a follow up PR.

event-exporter/kubernetes/podlabels/cache_collector.go
```diff
@@ -0,0 +1,159 @@
...
+	GetLabels(namespace, pod string) map[string]string
+}
+
+// podLabelCollectorWithCache implements PodLabelCollector interface with an internal pod label cache.
+type podLabelCollectorWithCache struct {
+	ignoredNamespaces     map[string]struct{}
+	client                kubernetes.Interface
+	getPodTimeout         time.Duration
+	cache                 *lru.Cache[cacheKey, map[string]string]
+	emptyLabelPodCache    *lru.Cache[cacheKey, time.Time] // cache the timestamp so we can check TTL.
```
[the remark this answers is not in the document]
leonzz: Yeah that's a good suggestion.    I think one cache with TTL that hold pods + labels (including empty labels) would simplify the code.
Will think about this and make follow up changes if necessary.

event-exporter/kubernetes/podlabels/cache_collector.go
```diff
@@ -0,0 +1,159 @@
...
+			return nil
+		} else {
+			recordEmptyLabelPodCacheExpire()
+			pc.emptyLabelPodCache.Remove(cacheKey{Namespace: namespaceName, Name: podName})
+		}
+	} else {
+		recordEmptyLabelPodCacheMiss()
+	}
+	ctx, cancel := context.WithTimeout(context.TODO(), pc.getPodTimeout)
+	pod, err := pc.client.CoreV1().Pods(namespaceName).Get(ctx, podName, metav1.GetOptions{})
```
CatherineF-dev, line 91: How to track this pc.client.CoreV1().Pods?

Not sure whether we already have other metrics to know event-exporter is calling pc.client.CoreV1().Pods x times per second.

If not, should we add a metric to track it?

CatherineF-dev:
Cool. 

Event-exporter memory is limited by the cache size.  Apiserver ddos `pod, err := pc.client.CoreV1().Pods(namespaceName).Get(ctx, podName, metav1.GetOptions{})` is not limited explicitly.  As long as we have metrics to monitor, we can root cause and mitigate quickly.

/lgtm

PR #867, opened 2025-04-09
cache GetMetricKind responses

### LINKED ISSUE #866: Hitting internal google quota "monitoring.googleapis.com/descriptor_read_requests"
We're seeing pretty consistant errors in some of our larger clusters where a lot of workloads are using pub/sub backlog scaling:

```
unable to get external metric pendo-io/pubsub.googleapis.com|subscription|num_undelivered_messages/&LabelSelector{MatchLabels:map[string]string{resource.labels.project_id: redacted,resource.labels.subscription_id: redacted,},MatchExpressions:[]LabelSelectorRequirement{},}: unable to fetch metrics from external metrics API: the server could not find the descriptor for metric pubsub.googleapis.com/subscription/num_undelivered_messages: googleapi: Error 429: Quota exceeded for quota metric 'Descriptor read requests' and limit 'Descriptor read requests per minute' of service 'monitoring.googleapis.com' for consumer 'project_number:redacted'.
Details:
[
  {
    "@type": "type.googleapis.com/google.rpc.ErrorInfo",
    "domain": "googleapis.com",
    "metadata": {
      "consumer": "projects/redacted",
      "quota_limit": "DescriptorReadRequestsPerMinutePerProject",
      "quota_limit_value": "6000",
      "quota_location": "global",
      "quota_metric": "monitoring.googleapis.com/descriptor_read_requests",
      "service": "monitoring.googleapis.com"
    },
    "reason": "RATE_LIMIT_EXCEEDED"
  }
]
, rateLimitExceeded
```

This seems to be coming from calls to `(t *Translator) GetMetricKind` which makes this call every time it's called

DP19:
fixes #866

This allows responses for `MetricDescriptors.Get` to be cached as to not hit the quota limit described in the linked issue


custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -509,9 +512,20 @@ func (t *Translator) ListMetricDescriptors(fallbackForContainerMetrics bool) *st
 		Filter(filter)
 }
 
+type cachedMetricInfo struct {
```
CatherineF-dev, line 515: replace it with LRUExpireCache https://pkg.go.dev/k8s.io/apimachinery/pkg/util/cache

DP19: Thanks, updated. I put some initial values here that i think would work. Let me know if we want to make this configurable.

CatherineF-dev: > Let me know if we want to make this configurable.

Yes, it will be better to make it configurable. And this feature can be turned off by default.

If we find more users are complaining this issue, we can make it turn on by default later.

CatherineF-dev: MetricKindCacheSize:
- When it's 0, it doesn't use the cache
- When it's not 0, it uses the cache

CatherineF-dev:
qq: is it ready for review?

DP19:
> qq: is it ready for review?

Yes this should be ready for eyes now. Thank you

DP19:
@CatherineF-dev I saw a similar implementation be put in place for caching External Metrics in #837 so i re-worked this PR to follow the same pattern. These should be fine to work together and there is a use case where you'd want to cache these GetMetricKind responses without caching the external metrics

custom-metrics-stackdriver-adapter/pkg/adapter/translator/query_builder.go
```diff
@@ -512,6 +520,14 @@ func (t *Translator) ListMetricDescriptors(fallbackForContainerMetrics bool) *st
 // GetMetricKind returns metricKind for metric metricName, obtained from Stackdriver Monitoring API.
 func (t *Translator) GetMetricKind(metricName string, metricSelector labels.Selector) (string, string, error) {
 	metricProj := t.config.Project
+	cacheKey := metricKindCacheKey{
```
juli4n, line 523: Could you add a test for GetMetricKind in query_builder_test.go that covers both cache hit and miss?

juli4n (commented):
Thanks for the contribution! Change looks good overall. Added some minor comments.

custom-metrics-stackdriver-adapter/adapter.go
```diff
@@ -222,6 +233,10 @@ func main() {
 		"The duration (e.g., 1m, 5s) for which external metric values are cached.")
 	flags.IntVar(&serverOptions.ExternalMetricsCacheSize, "external-metric-cache-size", serverOptions.ExternalMetricsCacheSize,
 		"The maximum number of entries in the external metric cache.")
+	flags.DurationVar(&serverOptions.MetricKindCacheTTL, "metric-kind-cache-ttl", serverOptions.MetricKindCacheTTL,
+		"The duration (e.g., 1m, 5s) for which metric kind info calls are cached.")
+	flags.IntVar(&serverOptions.MetricKindCacheSize, "metric-kind-cache-size", serverOptions.MetricKindCacheSize,
+		"The maximum number of entries in the metric kind cache. Defaults to 0 (off). For clusters using a lot of stackdriver based auto-scaling; adjusting this value to equal the separate metric types used will help reduce internal api rate limits")
```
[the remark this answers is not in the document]
DP19: Updated the language here; still not sure it's super clear. If you have suggestions let me know!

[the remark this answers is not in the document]
DP19: updated the new flags to indicate adding the TTL is what will enable/disable this feature

juli4n (approved):
LGTM. Thanks!

DP19:
@juli4n fixed the failing unit test 🙏🏻

Issue #480, opened 2022-07-15
Custom Metrics Adapter Applicability to GKE Autopilot


StevenACoffman:
In the tutorial https://cloud.google.com/kubernetes-engine/docs/tutorials/autoscaling-metrics#pubsub it is unclear whether the steps are applicable to an GKE Autopilot cluster.

For instance, is it necessary to install the custom-metrics-stackdriver-adapter in an GKE Autopilot cluster?
```
kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/k8s-stackdriver/master/custom-metrics-stackdriver-adapter/deploy/production/adapter_new_resource_model.yaml
```

StevenACoffman:
While I can see that in my GKE Autopilot clusters the `event-exporter-gke` pods are running in the `kube-system` namespace, there does not appear to be a pre-installed custom-metrics-stackdriver-adapter.

When installing in GKE Autopilot, you do not need need to, and in fact you cannot use node `iam.gke.io/gke-metadata-server-enabled: "true"`. Otherwise the directions for [how to Use Workload Identity are correct]( https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity)

See GoogleCloudPlatform/k8s-stackdriver#315 and GoogleCloudPlatform/kubernetes-engine-samples#140

```
gcloud iam service-accounts create custom-metrics-sd-adapter --project "$GCP_PROJECT_ID"

gcloud projects add-iam-policy-binding "$GCP_PROJECT_ID" \
  --member "serviceAccount:custom-metrics-sd-adapter@$GCP_PROJECT_ID.iam.gserviceaccount.com" \
  --role "roles/monitoring.editor"

gcloud iam service-accounts add-iam-policy-binding \
  --role roles/iam.workloadIdentityUser \
  --member "serviceAccount:$GCP_PROJECT_ID.svc.id.goog[custom-metrics/custom-metrics-stackdriver-adapter]" \
  "custom-metrics-sd-adapter@$GCP_PROJECT_ID.iam.gserviceaccount.com"

kubectl create -f https://raw.githubusercontent.com/GoogleCloudPlatform/k8s-stackdriver/master/custom-metrics-stackdriver-adapter/deploy/production/adapter.yaml

kubectl annotate serviceaccount custom-metrics-stackdriver-adapter \
[...]

PR #969, opened 2025-08-23
docs: clarify that stackdriver adapter is for end-users

juli4n:
This has caused some confusion among users (see https://github.com/GoogleCloudPlatform/k8s-stackdriver/issues/480#issuecomment-2493956681).

Fixes https://github.com/GoogleCloudPlatform/k8s-stackdriver/issues/480.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 > :exclamation: **Most tools in this repository are not meant for end-users.**
 > It contains source code for tools that are pre-installed in the GKE clusters.
 > An exception to this is the [Custom Metrics - Stackdriver Adapter](custom-metrics-stackdriver-adapter),
-> which is meant for end-user consumption.
+> which is meant for [end-user consumption][stackdriverAdapter].
 
 [Google Cloud Operations suite][cloudOperationsSite] (fka Stackdriver) provides advanced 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,9 +1,9 @@
 ## Google Cloud Operations integration for GKE
 
-> :exclamation: **Tools in this repository are not meant for 
-> end-users.**
-> It contains source code for tools that are pre-installed in 
-> the GKE clusters.
+> :exclamation: **Most tools in this repository are not meant for end-users.**
+> It contains source code for tools that are pre-installed in the GKE clusters.
+> An exception to this is the [Custom Metrics - Stackdriver Adapter](custom-metrics-stackdriver-adapter),
+> which is meant for end-user consumption.
```
rsgowman, line 6: Perhaps we should refer back to https://cloud.google.com/kubernetes-engine/docs/tutorials/autoscaling-metrics? That contains the instructions needed to actually make use of the cmsa.

(Mostly, I think we expect people to go from that page to here, rather than the other way around. But for anyone landing here first, having that context would be useful.)

lgtm either way though.

juli4n: Good idea. Done.

juli4n (commented):
Done.

PR #994, opened 2025-09-26
Event-exporter lister watch memory improvement.

zkdlin211:
Changes including:

1. Remove unused flag.
2. Change the flag `enableRemoveOptionsLimit` to `listerWatcherOptionsLimit` to allow configuration on number of options.limit. No limits when set to 0.
3. Support DeltaFIFO storage on event lister watcher.
4. Add flag `storageType` to allow configuring event-exporter lister watcher storage type.

### DIFF
--- a/event-exporter/Makefile
+++ b/event-exporter/Makefile
@@ -20,7 +20,7 @@ ALL_ARCH=amd64 arm64
 IMAGE_NAME = event-exporter
 
 PREFIX ?= staging-k8s.gcr.io
-TAG ?= v0.5.6
+TAG ?= v0.5.7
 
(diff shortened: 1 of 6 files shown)


event-exporter/Makefile
```diff
@@ -20,7 +20,7 @@ ALL_ARCH=amd64 arm64
 IMAGE_NAME = event-exporter
 
 PREFIX ?= staging-k8s.gcr.io
-TAG ?= v0.5.6
+TAG ?= v0.5.7
```
yungdarek122, line 23: nit: I think it is better to bump the version when releasing a new version

zkdlin211: I'll bump the version in the same PR.

PR #1014, opened 2025-11-18
Generate mutiplatform manifests for kubelet-to-gcm

nikhilnishad-goog:
Create manifests that support multiple platforms for the kubelet-to-gcm integration.

platforms = linux/arm64 (new) and linux/amd64 only for now.


nikhilnishad-goog:
Fixes #1014

baizhenyu (approved):

PR #1015, opened 2025-11-19
Prioritize the high-precision timestamp from the newer events.k8s.io/v1 schema.

zkdlin211:
Prioritize the high-precision timestamp from the newer events.k8s.io/v1 schema.


event-exporter/sinks/stackdriver/log_entry_factory.go
yungdarek122: Just to be safe, can we have a flag guarding this change?

PR #1025, opened 2025-12-12
Move -$$arch suffix from image name to tag. Bump kubelet-to-gcm version to 1.4.4

jakweg:
By making images have the same name, but different tags make it easier for Google internal infrastructure to build them. 

It shouldn't affect any existing users of the image, as there will be a docker manifest pushed in place of the old image.

### DIFF
--- a/kubelet-to-gcm/Makefile
+++ b/kubelet-to-gcm/Makefile
@@ -41,7 +41,6 @@ clean:
 	rm -rf build
 
 .container:
-	echo $(IMAGE):$(TAG)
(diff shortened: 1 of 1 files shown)


baizhenyu:
You will need a new release for this change, do you want to bump the version?

jakweg:
Bumped the version (bumped patch version only).

baizhenyu:
Have you tested this Makefile locally?

jakweg:
Yes, tested locally

baizhenyu (approved):

PR #1040, opened 2026-01-13
Add restart attempt and uid labels for jobset events

chelseychen:
This PR extends the pod label extraction logic in the Event Exporter to support the `jobset.sigs.k8s.io/restart-attempt` and `jobset.sigs.k8s.io/jobset-uid` labels.

Previously, we only extracted the `jobset_name` for pod owner labels. By adding the `restart_attempt` and `uid`, we allow downstream consumers to clearly distinguish between events occurring across different restart attempts of the same JobSet. This is critical for accurate reliability metrics and debugging restart loops where events might overlap in time.

### DIFF
--- a/event-exporter/kubernetes/podlabels/label_utils.go
+++ b/event-exporter/kubernetes/podlabels/label_utils.go
@@ -5,9 +5,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


chelseychen:
Fixes #1041

PR #1043, opened 2026-01-15
Enable streaming listWatcher only if the required k8s feature gate is enabled

zkdlin211:


event-exporter/kubernetes/watchers/events/watcher.go
```diff
@@ -174,3 +178,22 @@ eventLoop:
...
+		glog.Errorf("fail to get raw metrics from the API server: %v, fall back to false", err)
+		return false
+	}
+
+	// Look for: kubernetes_feature_enabled{name="WatchList",stage="BETA"} 1
+	metricLine := "kubernetes_feature_enabled{name=\"" + featureName + "\""
+	lines := strings.Split(string(data), "\n")
+	for _, line := range lines {
+		if strings.Contains(line, metricLine) {
+			return strings.HasSuffix(line, "1")
```
yungdarek122: Do we need to worry the value is expressed as 1.0 instead of 1?

zkdlin211: added check

PR #1045, opened 2026-01-20
update extractAllLabels to filter out labels that aren't metric labels

jinyigke:


jinyigoogle:
> Please add unit tests.
> 
> Also bump the version in https://github.com/GoogleCloudPlatform/k8s-stackdriver/blob/master/prometheus-to-sd/Makefile#L21

Done

PR #1081, opened 2026-03-04
custom-metrics-stackdriver-adapter: Fix GPU/TPU tolerations for GKE Autopilot compatibility

courageJ:
#1078 added tolerations to allow scheduling to GPU/TPU nodes if they are the only type of nodes. However, GKE Autopilot rejects such tolerations if they don't request gpu/tpu resources. 

This is a workaround to bypass GKE Autopilot wehbook. We will follow up with GKE Autopilot to allowlist this component.


erain (approved):
LGTM

PR #1186, opened 2026-05-12
Tighten resource attribution in the Stackdriver event sink

erain:
## Summary

When the event-exporter maps a `corev1.Event` onto a Stackdriver `MonitoredResource`, the Pod/Node labels are taken from the event's `involvedObject` reference. These reference fields are not cross-checked against the event's own `metadata.namespace`, which is the value the API server validates when an event is created.

This change adds that cross-check inside `resourceFromEvent`:

- **Pod events** must satisfy `event.Namespace != \"\"` and `event.Namespace == event.InvolvedObject.Namespace`. When the check passes, `namespace_name` is sourced from `event.Namespace` directly. Otherwise the entry falls back to the default cluster-scoped resource.
- **Node events** (which have no per-event namespace to compare against) are attributed to `k8s_node` only when the event originates from one of `default`, `kube-system`, or `kube-node-lease`. Otherwise the entry falls back to the default cluster-scoped resource.

The fallback preserves the log content under the cluster resource, so events that don't pass the check are still exported and inspectable; they just stop being attributed to the workload they reference.

## Test plan

- [x] `go test -mod=vendor ./...` in `event-exporter/` is green.
- [x] `monitored_resource_factory_test.go` gains coverage for: pod event whose event namespace disagrees with the involved object, pod event with empty event namespace, node event in trusted namespace, node event in untrusted namespace.
- [x] `log_entry_factory_test.go` cases updated to populate `event.Namespace` consistent with the new requirement; pod-label enrichment path still exercised end-to-end.

### DIFF
--- a/event-exporter/sinks/stackdriver/log_entry_factory_test.go
+++ b/event-exporter/sinks/stackdriver/log_entry_factory_test.go
@@ -111,6 +111,7 @@ func TestFromEvent(t *testing.T) {
 		{
 			desc: "k8s pod event with pod labels",
 			event: &corev1.Event{
+				ObjectMeta:     metav1.ObjectMeta{Namespace: "test_namespace"},
 				Type:           "Normal",
 				InvolvedObject: involvedPodObject,
 				LastTimestamp:  lastTimestamp,
--- a/event-exporter/sinks/stackdriver/monitored_resource_factory.go
+++ b/event-exporter/sinks/stackdriver/monitored_resource_factory.go
@@ -74,7 +74,15 @@ func (f *monitoredResourceFactory) resourceFromEvent(event *corev1.Event) *sd.Mo
 
 	switch event.InvolvedObject.Kind {
 	case pod:
-		monitoredResource = f.buildPodMonitoredResource(event)
+		// The event's own metadata.namespace is the RBAC-enforced source
+		// of truth for where the event was created. Only emit a pod-scoped
+		// resource when the involved object's namespace agrees with it;
+		// otherwise fall back to the default cluster-scoped resource.
+		if event.Namespace != "" && event.Namespace == event.InvolvedObject.Namespace {
+			monitoredResource = f.buildPodMonitoredResource(event)
+		} else {
+			monitoredResource = f.defaultResource
+		}
 	case node:
 		monitoredResource = f.buildNodeMonitoredResource(event)
 	default:
@@ -86,7 +94,7 @@ func (f *monitoredResourceFactory) resourceFromEvent(event *corev1.Event) *sd.Mo
 func (f *monitoredResourceFactory) buildPodMonitoredResource(event *corev1.Event) *sd.MonitoredResource {
 	labels := copyMap(f.commonLabels)
 	labels[podName] = event.InvolvedObject.Name
-	labels[namespaceName] = event.InvolvedObject.Namespace
+	labels[namespaceName] = event.Namespace
 
 	return &sd.MonitoredResource{
 		Type:   k8sPod,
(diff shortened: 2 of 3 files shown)


erain:
Pushed `f85462c` to address the review feedback:

- Dropped `default` from the trusted set in `monitored_resource_factory.go`; only `kube-system` and `kube-node-lease` now produce a node-scoped `MonitoredResource`. Pod attribution is unchanged (still requires `event.Namespace != ""` and `event.Namespace == event.InvolvedObject.Namespace`, with `namespace_name` sourced from `event.Namespace`).
- `monitored_resource_factory_test.go`: the previous `default` → `k8s_node` case now asserts the cluster fallback; a `kube-node-lease` → `k8s_node` case is added so both trusted namespaces are covered.
- `log_entry_factory_test.go`: node fixtures switched from `default` to `kube-system` so the trusted-namespace path is still exercised.

`go test -mod=vendor ./...` from `event-exporter/`:

```
?   	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter	[no test files]
ok  	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/kubernetes/podlabels	(cached)
?   	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/kubernetes/watchers	[no test files]
ok  	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/kubernetes/watchers/events	(cached)
?   	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/sinks	[no test files]
ok  	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/sinks/stackdriver	0.820s
?   	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/utils	[no test files]
```

erain:
Force-pushed `208ca39` to scope this PR strictly to Pod attribution:

- Removed `trustedNodeEventNamespaces` and reverted `resourceFromEvent` so the `Node` case goes straight to `buildNodeMonitoredResource(event)` exactly as on master. Node attribution will be addressed in a follow-up PR.
- Kept the Pod fix: `k8s_pod` is emitted only when `event.Namespace != ""` and `event.Namespace == event.InvolvedObject.Namespace`; `namespace_name` is sourced from `event.Namespace`; otherwise fall back to the cluster resource.
- `monitored_resource_factory_test.go`: keeps the matching-namespace pod case, adds mismatched-namespace and empty-event-namespace fallback cases; node-allowlist cases removed.
- `log_entry_factory_test.go`: only the pod fixture gains `ObjectMeta.Namespace`; node fixtures are back to their original (no namespace) form.

`go test -mod=vendor ./...` from `event-exporter/`:

```
?   	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter	[no test files]
ok  	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/kubernetes/podlabels	(cached)
?   	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/kubernetes/watchers	[no test files]
ok  	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/kubernetes/watchers/events	(cached)
?   	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/sinks	[no test files]
ok  	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/sinks/stackdriver	0.829s
?   	github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter/utils	[no test files]
```

courageJ (approved):

PR #1197, opened 2026-05-20
Prevent http.DefaultServeMux exposure on CMSA

laoj2:
When the Prometheus /metrics endpoint is enabled, the HTTP server was initialized via `http.ListenAndServe(addr, nil)`, defaulting to the global `http.DefaultServeMux`

Transitive third-party dependencies (like k8s libraries) register global debug and profiling handlers (`/debug/pprof/*`), these  endpoints were unintentionally exposed publicly on the metrics port.

So, we're isolating the `/metrics` endpoint by initializing a dedicated `ServeMux`.

Tested on a live cluster with `--metrics-address=0.0.0.0:8080` flag configured on the CMSA deployment:

Metrics scraping still works:

```
curl -i http://localhost:8080/metrics
HTTP/1.1 200 OK
```

Profiling endpoints are no longer exposed as a side effect:

```
curl -i http://localhost:8080/debug/pprof/
HTTP/1.1 404 Not Found 
```

### DIFF
--- a/custom-metrics-stackdriver-adapter/adapter.go
+++ b/custom-metrics-stackdriver-adapter/adapter.go
@@ -291,8 +291,9 @@ func getDefinitionName(namer *openapinamer.DefinitionNamer, name string) (string
 }
 
 func runPrometheusMetricsServer(addr string) {
-	http.Handle("/metrics", promhttp.Handler())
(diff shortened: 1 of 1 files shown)


rsgowman (approved):
lgtm.

PR #1198, opened 2026-05-20
Hardening URL Construction and IPv6 Support

courageJ:
Hardens URL construction for Kubelet and Controller manager metrics endpoints by replacing string interpolation with `net.JoinHostPort`. This ensures correct handling of IPv6 addresses (bracketing) and improves overall robustness.

### Changes:
- **controller/client.go**: Use `net.JoinHostPort` for robust `/metrics` endpoint construction.
- **kubelet/client.go**: Use `net.JoinHostPort` for robust `/stats/summary` endpoint construction.
- **controller/client_test.go**: Added unit tests for `NewClient` (v4, v6, hostname).
- **kubelet/client_test.go**: Added unit tests for `NewClient` (HTTP/S, v4/v6).

### Testing:
- All new tests passed for both packages.
- Verified correct bracketing for IPv6 in both clients via unit tests.

### DIFF
--- a/kubelet-to-gcm/monitor/controller/client.go
+++ b/kubelet-to-gcm/monitor/controller/client.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -87,7 +88,7 @@ type Client struct {
 // NewClient generates a client to hit the given controller.
 func NewClient(host string, port uint, client *http.Client) (*Client, error) {
 	// Parse our URL upfront, so we can fail fast.
-	urlStr := fmt.Sprintf("http://%s:%d/metrics", host, port)
+	urlStr := fmt.Sprintf("http://%s/metrics", net.JoinHostPort(util.NormalizeHost(host), fmt.Sprint(port)))
 	metricsURL, err := url.Parse(urlStr)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


erain (changes requested):
This fixes raw IPv6 literals, but it regresses the bracketed IPv6 form that users would have needed as the existing workaround, for example `--kubelet-host=[2001:db8::1]`. Passing that value to `net.JoinHostPort` double-brackets the host, producing an invalid URL. Since this flag is user-provided and the old implementation accepted bracketed IPv6, the hardening should normalize that input before joining.

Please strip one surrounding IPv6 bracket pair before calling `net.JoinHostPort` or add a small host-normalization helper used by both clients. Add tests for bracketed IPv6 hosts in both `monitor/controller` and `monitor/kubelet` so this does not regress again.

PROMPT for coding agent: Fix PR #1198 by adding shared host normalization for `NewClient` URL construction. It should accept raw IPv6, already-bracketed IPv6, IPv4, and hostnames, then call `net.JoinHostPort` with the normalized host. Update controller and kubelet client tests to cover bracketed IPv6 in addition to the existing raw IPv6 cases.

erain (approved):
Re-reviewed the updated URL construction change. `NormalizeHost` now handles already-bracketed IPv6 before `net.JoinHostPort`, and the controller/kubelet tests cover raw and bracketed IPv6 cases. Ran `go test ./monitor/controller ./monitor/kubelet ./monitor/util` from `kubelet-to-gcm` on the PR head. Looks good.

PR #1201, opened 2026-05-21
Enforce response body limits in kubelet-to-gcm

courageJ:
This PR implements size limits when reading HTTP response bodies in the kubelet-to-gcm monitoring daemon to prevent potential unbounded memory allocation.

Key changes:
- Added io.LimitReader to getGCEMetaData, doRequestAndParse, and doRequestAndUnmarshal.
- Set specific limits: 10MB for GCE metadata and Prometheus metrics, 50MB for Kubelet stats summaries.
- Replaced deprecated ioutil.ReadAll with io.ReadAll.
- Added unit tests in monitor/controller and monitor/kubelet to verify size limiting behavior and ensure success cases continue to work.

Tested locally with go test ./monitor/... in kubelet-to-gcm.

### DIFF
--- a/kubelet-to-gcm/monitor/config/initialize.go
+++ b/kubelet-to-gcm/monitor/config/initialize.go
@@ -18,17 +18,20 @@ package config
 
 import (
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"strings"
 	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-stackdriver/kubelet-to-gcm/monitor"
+	"github.com/GoogleCloudPlatform/k8s-stackdriver/kubelet-to-gcm/monitor/util"
 )
 
 const (
 	gceMetaDataEndpoint = "http://169.254.169.254"
 	gceMetaDataPrefix   = "/computeMetadata/v1"
+	// maxResponseBodySize is the maximum size of the response body we will read.
+	// 10MB is more than enough for GCE metadata responses.
+	maxResponseBodySize = 10 * 1024 * 1024
 )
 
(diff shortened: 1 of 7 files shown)


erain (changes requested):
The implementation caps allocation, but it does not actually enforce the response-size limit. `io.LimitReader` returns EOF after `maxResponseBodySize`, so callers cannot tell whether the server sent exactly the allowed size or much more. That can silently accept truncated GCE metadata and can silently drop controller metrics if the first limited chunk is parseable.

Please read at most `maxResponseBodySize+1` bytes or use an `io.LimitedReader` and explicitly return an error when the body exceeds the configured limit. Apply the same helper to `monitor/config/initialize.go`, `monitor/controller/client.go`, and `monitor/kubelet/client.go`. The tests should include oversized-but-otherwise-valid responses and assert the explicit size-limit error, not just parse failures from invalid truncated data.

PROMPT for coding agent: Fix PR #1201 by replacing every `io.ReadAll(io.LimitReader(..., maxResponseBodySize))` call with a shared helper that detects bodies larger than the limit and returns an error. Use it for GCE metadata, controller metrics, and kubelet summaries. Update or add unit tests so oversized valid payloads fail with the size-limit error, while under-limit payloads still pass.

erain (approved):
Re-reviewed the updated response-body limit implementation. The new shared `util.ReadWithLimit` reads `limit+1` bytes and returns an explicit `ErrBodyTooLarge`, which addresses the previous truncation concern, and the callers in metadata, controller, and kubelet paths now use it. Ran `go test ./monitor/...` from `kubelet-to-gcm` on the PR head. Looks good.

PR #1203, opened 2026-05-21
Fix potential nil pointer dereferences in kubelet-to-gcm

courageJ:
This PR hardens the `kubelet-to-gcm` monitoring daemon against crashes by adding nil pointer protections across several components.

### Changes

- **Translation Logic**: Added nil checks for optional Kubelet stats fields (`UsedBytes` in `FsStats`, `MajorPageFaults` in `MemoryStats`) before dereferencing them in `monitor/kubelet/translate.go`.
- **Service & Client Safety**: Added nil checks for the GCM service pointer in the `Once` function (`monitor/poll.go`) and for HTTP responses in client helpers (`monitor/controller/client.go` and `monitor/kubelet/client.go`) to prevent panics during `defer response.Body.Close()`.
- **Helpers**: Added `Uint64Ptr` to the `monitor` package for parity with existing pointer helpers.
- **Testing**: Added a `TestTranslator_NilFields` regression test in `monitor/kubelet/translate_test.go` to verify that missing optional fields in `stats.Summary` do not cause panics.

### Verification

- All existing unit tests passed.
- New regression test `TestTranslator_NilFields` passed and verified the fixes.

### DIFF
--- a/kubelet-to-gcm/monitor/controller/client.go
+++ b/kubelet-to-gcm/monitor/controller/client.go
@@ -98,6 +98,9 @@ func (c *Client) doRequestAndParse(req *http.Request) (*Metrics, error) {
 	if err != nil {
 		return nil, err
 	}
+	if response == nil {
+		return nil, fmt.Errorf("empty response from controller")
+	}
 	defer response.Body.Close()
 	body, err := ioutil.ReadAll(response.Body)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


erain (approved):
Reviewed the nil pointer hardening changes and ran `go test ./monitor/...` from `kubelet-to-gcm` on the PR head. The new guards cover the panic cases without changing the existing required-field behavior outside those optional fields. Looks good.

PR #1231, opened 2026-07-22
Sanitize Stackdriver API error responses to prevent SA email disclosu…

sophieliu15:
Prevent raw underlying Stackdriver API error details from being formatted directly into client-facing StatusError messages when looking up metric descriptors.

Previously, errors returned by the Stackdriver API (such as 403 Forbidden responses when querying inaccessible projects) were formatted directly into the HTTP response body returned to API callers, potentially exposing underlying GCP Service Account email identities.

This change logs the detailed error internally via klog for server-side debugging while returning a generic metric-not-found StatusError message to the client. Unit tests are also added to verify error sanitization.

### DIFF
--- a/custom-metrics-stackdriver-adapter/pkg/adapter/translator/errors_test.go
+++ b/custom-metrics-stackdriver-adapter/pkg/adapter/translator/errors_test.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


custom-metrics-stackdriver-adapter/pkg/adapter/translator/errors_test.go
```diff
@@ -0,0 +1,48 @@
...
+	if statusErr == nil {
+		t.Fatalf("expected non-nil StatusError")
+	}
+
+	if statusErr.ErrStatus.Code != http.StatusNotFound {
+		t.Errorf("expected status code %d, got %d", http.StatusNotFound, statusErr.ErrStatus.Code)
+	}
+
+	expectedMsg := "the server could not find the descriptor for metric custom.googleapis.com/test_metric"
+	if statusErr.ErrStatus.Message != expectedMsg {
```
laoj2: what about removing this check? (given that we're checking if the status message isn't leaking the sensitive error below)

Checking for the exact status error message can be a change-detector check.

laoj2 (approved):

custom-metrics-stackdriver-adapter/pkg/adapter/translator/errors_test.go
sophieliu15: Done

PR #1244, opened 2026-08-24
Add opt-in pprof HTTP server for profiling

CatherineF-dev:
Expose Go's net/http/pprof handlers on a dedicated HTTP server, enabled only when the PPROF_SERVER_PORT environment variable is set (disabled by default). The server runs on its own ServeMux and port, is started in a goroutine, and is gracefully shut down when the process stop channel is closed.

Move the Prometheus /metrics handler off http.DefaultServeMux onto a private ServeMux so that importing net/http/pprof (which registers its handlers on the default mux in its init) cannot inadvertently expose the profiling endpoints on the metrics port.

Add a unit test for startPprofServer.

### DIFF
--- a/event-exporter/Makefile
+++ b/event-exporter/Makefile
@@ -20,7 +20,7 @@ ALL_ARCH=amd64 arm64
 IMAGE_NAME = event-exporter
 
 PREFIX ?= staging-k8s.gcr.io
-TAG ?= v0.5.11
+TAG ?= v0.5.13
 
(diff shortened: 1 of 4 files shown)


baizhenyu (approved):
/lgtm Only one minor thing, this is the first instance that using env var for a config (i.e other configs are passed by CLI flags). Consider making them consistent if feasible.

CatherineF-dev:
I verified that the newly added env is not reconciled.

